Issue #56, opened 2014-12-09
WindowsError in removing temp files


lapsed-pacifist:
```python
WindowsError: [Error 2] The system cannot find the file specified: 'C:\\Python27\\lib\\site-packages\\gooey-0.1.0-
py2.7.egg\\gooey\\python_bindings\\da56848393418e7c55974af7edbbb2d2.pyc'
```
This just results from the inclusion of the decorator in any program

ardoramor:
I'm getting the same error:

...
    main()
  File "C:\Python27\lib\site-packages\gooey-0.1.0-py2.7.egg\gooey\python_bindings\gooey_decorator.py", line 119, in inner
    parser = get_parser(main_module_path)
  File "C:\Python27\lib\site-packages\gooey-0.1.0-py2.7.egg\gooey\python_bindings\gooey_decorator.py", line 162, in get_parser
    return source_parser.extract_parser(module_path)
  File "C:\Python27\lib\site-packages\gooey-0.1.0-py2.7.egg\gooey\python_bindings\source_parser.py", line 145, in extract_parser
    client_module = modules.load(module_source)
  File "C:\Python27\lib\site-packages\gooey-0.1.0-py2.7.egg\gooey\python_bindings\modules.py", line 33, in load
    os.remove(tmp_pyc_file)
WindowsError: [Error 2] The system cannot find the file specified: 'C:\\Python27\\lib\\site-packages\\gooey-0.1.0-py2.7.egg\\gooey\\python_bindings\\e68dbae79cf81af7bd68c8d4b59c07e.pyc'

I have Python 2.7 installed. Installed master Gooey. Seems like it has problem compiling binding?

chriskiehl:
Could you post the code you were attempting to run? Stacktrace alone doesn't give quite enough info.

ardoramor:
import argparse
from gooey import Gooey

@Gooey
def main():
	parser = argparse.ArgumentParser(description='UV Metadata Uploader.')
	parser.add_argument('server_url', 
			help='URL of the UV server where the metadata will be uploaded.')
	parser.add_argument('api_key', 
			help='Server API key that authenticates to an account with permissions to manage metadata of a feed.')
	parser.add_argument('feed_id', type=int, nargs=1, 
			help='ID of the feed for which the metadata will be uploaded.')
	parser.add_argument('gpx_files', type=argparse.FileType('r'), nargs='+', 
			help='File containing GPX metadata that will be uploaded to the UV server.')
[...]

PR #62, opened 2015-01-15
fix indentation in returned parser source #56

aphorton:
This fixes a bug in source_parser.py, where main() was assumed to be indented with spaces.


chriskiehl:
Thanks for the fix!

PR #73, opened 2015-03-06
README: "Radio Group" => RadioGroup

cvrebert:
Remove errant space in widget name


chriskiehl:
Take *that*, errant space!

PR #77, opened 2015-03-17
Various fixes and AutoGooey

lrq3000:
Various fixes to make Gooey more usable. Some changes were necessary to make Gooey work with my script (which supplies a custom argv to argparse).

I also added an entry in the README about a small snippet called "AutoGooey decorator" so that an application can both be used with Gooey and without.

A more elegant solution would be to provide this functionality in Gooey itself, as an additional argument of the gooey.Gooey() decorator function:

    def Gooey(f=None, advanced=True,
            language='english', show_config=True,
            program_name=None, program_description=None,
            enable_on=None):

        if enable_on and len(sys.argv) > 1 and sys.argv[1] != enable_on:
            return f
        else: # Avoid infinite recursion calls to Gooey decorator by replacing the enable_on argument
            sys.argv[1] = sys.argv[1] + "_done"
        # ... rest of the decorator ...

This would for example allow to use @Gooey(enable_on="--gui") to enable the decorator only when --gui is given as an argument.

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,8 +187,13 @@ Note: notice that you still need to decorate your function with @Gooey.
 
 | Widget         |           Example            | 
 |----------------|------------------------------| 
-|  Directory/FileChooser   | <p align="center"><img src="https://raw.githubusercontent.com/chriskiehl/Gooey/master/resources/filechooser.gif" width="400"></p> | 
+|  FileChooser<br />FileSaver<br />DirChooser<br />MultiFileChooser   | <p align="center"><img src="https://raw.githubusercontent.com/chriskiehl/Gooey/master/resources/filechooser.gif" width="400"></p> | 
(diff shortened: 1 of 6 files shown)


lrq3000:
Note to @chriskiehl : All the commits are clean normally except maybe one:

https://github.com/lrq3000/Gooey/commit/0fb81e2bb9a6f85c7938bfcb23944b15ad1e54b8

This is an important fix for optional arguments to work, because otherwise Gooey will always specify all arguments (including optionals) to argparse with empty values, and in my setup this produced a buggy behavior (argparse would spit an error). I tested this commit and it seems that Gooey works correctly in all other aspects, but since you better know the codebase, you should check if I didn't do useless changes somewhere (you can try with my pyFileFixity python scripts to check if the error appears again).

chriskiehl:
I actually pulled that commit in last night. It did introduce a small quirk with positional arguments (returning None instead of the actual value), but other than that it resolves the issue perfectly! I'll push that merge and a few other changes tonight.  Thanks for tracking down the errors!

lrq3000:
Hey there, I just noticed that this pull request is still open. I noticed that you didn't merge my changes to the Readme (introducing AutoGooey to have both GUI and command-line functionality) and the update about custom widgets types table. Is that on purpose?

About AutoGooey, if it's not to be put into the Readme, you may put it in the new GooeyExamples, because the AutoGooey cannot be directly included in the codebase of Gooey (at least I couldn't achieve it, the AutoGooey decorator needs to be on the same script as where it is meant to be used, it cannot be put into a module and imported, this produces an error).

lrq3000:
BTW I'm very glad to see that this project is still living, I hope it will continue for a long time :) The move to make it language agnostic is very interesting, and it may bring this great tool to a great deal wider audience.

chriskiehl:
Hey @lrq3000, 

Sorry for leaving you hanging! Your idea of using a command line flag to turn Gooey on and off was a really good one -- in fact, as of tonight it's been rolled into the project (turned out to be the perfect solution to that import bug! (thanks!)). 

I didn't pull in your doc changes initially as I wasn't a fan of having a recipe section. My main goal with Gooey has always been simplicity above all else, so I didn't want something that should have been a feature of the project to fall onto the user to implement. 


So, I'm (finally) closing this now. Doc changes didn't make it, but the feature itself did! : )

PR #90, opened 2015-06-04
Multi dir dialog

Roshgar:
Hello, 
So, i appreciated a lot this nifty little module and also wanted something like this for a while now but couldnt find a simple, neatly packaged module. Gooey is the one!

I had a go at trying to implement a multiDirDialog. So it is functionnal, up to you if you wish to have it or not and if you find a better way of implementing it. :-)

Used the `MultiDirDialog` from wx.lib.agw and chucked in some formatting of the return value (For some weird reason it explicitly states that it is a `"Local Disk (C:)"` Instead of just `"C:"` when returning the path)

Anyhow, i'm also relatively new to actually doing something with a project (I.e putting it online so other people can glitch and not only you!) So if there are any procedures to be done before this pull request is created, sorry, I probably did not do it. (Other than testing a few cases.)

Please, just don't have a weird username like on one of my machines (ðø/any unicode type of characters don't pass well.) and it shouldn't have a problem.

Only three issues i can think about are : 
1 - The weird format of the path when MDD returns. I.E : `"Local Disk (C:)/Users/User/Documents"`
2 - If you happen to have accents/unicode characters in your username/Directory names, it freaks out at `return '{0} "{1}"'.format(self.option_string, value)`
3 - If you happen to manage to select a non existing directory (Shouldn't be possible but if you decide to remove the `MDD.DD_DIR_MUST_EXIST` it will be), AND that directory is somehow not created with the MDD dialog (If you play around with that code too) AND the error message stating `the directory does not exist` contains accents on your system, it will freak out due to unicode at `i18n._('uh_oh').format(error_msg)`

### DIFF
--- a/gooey/gui/widgets/widget_pack.py
+++ b/gooey/gui/widgets/widget_pack.py
@@ -93,7 +93,6 @@ class BaseFileChooser(BaseChooser):
       self.text_box.SetValue(result)
 
   def get_path(self, dlg):
-    print("In get_path")
     if isinstance(dlg, wx.DirDialog):
       return dlg.GetPath()
     else:
--- a/gooey/gui/widgets/widget_pack.py
+++ b/gooey/gui/widgets/widget_pack.py
@@ -93,51 +93,12 @@ class BaseFileChooser(BaseChooser):
       self.text_box.SetValue(result)
 
   def get_path(self, dlg):
+    print("In get_path")
     if isinstance(dlg, wx.DirDialog):
       return dlg.GetPath()
     else:
       paths = dlg.GetPaths()
       return paths[0] if len(paths) < 2 else ' '.join(paths)
-
-
-def correctPath(path): # This had to be made because it was returning a weird ass Path looking like : Local Disk (C:)/Users/..../Folder
-  tab = path.split('\\')
-  disk = re.split("[()]", tab[0])[1]
-  tab.pop(0)
-  for line in tab:
-    disk += u"/" + line
-  return (disk)
-
-def correctPaths(myList): # Cf correctPath
-  newList = []
-  for field in myList:
-    if not os.path.exists(field): # If it returns a "Local Disk (C:)/.../Folder" Bullshit.
-      newList.append(correctPath(field))
-    else:
-      newList.append(field)
-  return (newList)
-
-
-
-class BaseMultiDirChooser(BaseChooser): # kind of just copied yours.
-  def __init__(self, dialog): # Same code here normally.
-    BaseChooser.__init__(self)
-    self.dialog = dialog
-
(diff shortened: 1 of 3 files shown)


gooey/gui/widgets/widget_pack.py
```diff
@@ -97,6 +100,45 @@ def get_path(self, dlg):
       return paths[0] if len(paths) < 2 else ' '.join(paths)
 
 
+def correctPath(path): # This had to be made because it was returning a weird ass Path looking like : Local Disk (C:)/Users/..../Folder
+  tab = path.split('\\')
+  disk = re.split("[()]", tmpTab)[1]
```
chriskiehl: Left in some debug code? : ) No "tmpTab" is defined.

Roshgar:
Wooops. Didn't push the last commit. Sorry about that. :-)

chriskiehl:
I think there may be an easier solution. Check out the discussion [here](http://wxpython-dev.wxwidgets.narkive.com/AqaNgKLM/wx-lib-agw-multidirdialog-path-retrieval-issue). 

The short version is that If you pass off the `GetPaths` functionality to the underlying `GenericDirCtrl`, it will generate the correct paths for you. That way you don't have to do any manual path cleaning on your own (which wasn't working on linux). 

I believe something like this should work: 

    class MyMutiDirDlg(MultiDirDialog): 
        # some __init__ code 
        
        def GetPaths(self, *args, **kwargs): 
            # call the GenericDirCtrl instance on the parent
            return self.dirCtrl.GetPaths()

Roshgar:
Indeed. Much more simplified method. Tested it out works brilliantly (Also keeps the same window style as the fileSelector. So preferable) Will get back to that when either i make gooey compile or when i decide to give up and script a setup.

Roshgar:
And now i believe it should be back to a strict minimum of lines added/modified. (and functionnal at least on both my machines)

Roshgar:
Ok. I was either very tired or entirely elsewhere when i last tested this. Needed an extra modification in the wx directory that i had forgotten about. But that also means it's technically not fusable with Gooey. Bah. Too bad. I'll have my local version and install via script and cp's. Thanks for the project! (really do like it)

Over and out.

chriskiehl:
Wait, what happened? You needed to modify wx's core code?

Roshgar:
Oh god... I need to sleep and drink less coffee...
Yeah. I have no idea why but i was thinking private attributes. So i went
into wx's source code to do a dirCtr.GetPaths(). Instead of simply doing a
dlg.dirCtrl.Getpaths() from the gooey code.
Shoot me.
Wrapping it up right now and sending it over as a different pull request.

PR #91, opened 2015-06-12
MDD added

Roshgar:
Trashed my old version. Took your last up to date version, added only the necessary.
Should now be good.

### DIFF
--- a/gooey/gui/widgets/components.py
+++ b/gooey/gui/widgets/components.py
@@ -232,5 +232,5 @@ DateChooser       = lambda data: BaseGuiComponent(data=data, widget_pack=widget_
 TextField         = lambda data: BaseGuiComponent(data=data, widget_pack=widget_pack.TextInputPayload())
 Dropdown          = lambda data: BaseGuiComponent(data=data, widget_pack=widget_pack.DropdownPayload())
 Counter           = lambda data: BaseGuiComponent(data=data, widget_pack=widget_pack.CounterPayload())
-MultiDirChooser   = lambda data: BaseGuiComponent(data=data, widget_pack=widget_pack.MultiDirPayload())
+MultiDirChooser   = lambda data: BaseGuiComponent(data=data, widget_pack=widget_pack.MultiDirChooserPayload())
 
--- a/gooey/gui/widgets/components.py
+++ b/gooey/gui/widgets/components.py
@@ -232,4 +232,5 @@ DateChooser       = lambda data: BaseGuiComponent(data=data, widget_pack=widget_
 TextField         = lambda data: BaseGuiComponent(data=data, widget_pack=widget_pack.TextInputPayload())
 Dropdown          = lambda data: BaseGuiComponent(data=data, widget_pack=widget_pack.DropdownPayload())
 Counter           = lambda data: BaseGuiComponent(data=data, widget_pack=widget_pack.CounterPayload())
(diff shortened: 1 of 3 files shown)


Roshgar:
I'm also quite tempted to put back the `';'.join(paths)` instead of the current `' '.join(paths)` seeing as there will be spaces in file/directory names.

chriskiehl:
>seeing as there will be spaces in file/directory names. 

You actually raise a good point about spaces in the file path. early on, there were several issues in the tracker about wrapping file paths in quotes. We should repeat that pattern here rather than introducing a new delimiter. 

Some code notes: 

You've got a few DRY violations in there : ) I think there are two ways we could approach 'em. 

Option 1: Keep the `MultiDirChooser` type that you've created. You've copied some of the methods from `BaseFileChooser` verbatim to your class. Rather, than duplicating that code, I think you could just inherit from that base `BaseFileChooser` type, and then override only the methods that are different (e.g. `get_path`). That way you don't have the same code in two places. 


    class MultiDirChooser(BaseFileChooser):
      def __init__(self, *args, **kwargs):
        super(multiDirChooser, self).__init__(*args, **kwargs)
  
      def get_path(self, dlg):
        paths = dlg.dirCtrl.GetPaths()
        return paths[0] if len(paths) < 2 else ' '.join(paths)


Option 2: 

Modify `MDD.MultiDirDialog`. You can actually reuse all of `BaseFileChooser`'s code without modification if you override `MultiDirDialog`s implementation. If you do that, then it'll behave like all of the other `wx` dialogs and everything should gel together nicely. 

    class MyMultiDirChooser(MDD.MultiDirDialog):
      def __init__(self, *args, **kwargs):
        super(MyMultiDirChooser,self).__init__(*args, **kwargs)

      def GetPaths(self):
        return self.dirCtrl.GetPaths()

One small benefit is that the declaration pattern at the bottom of the file would be a bit more uniform: 

    MultiDirChooserPayload = partial(BaseFileChooser, dialog=lambda parent: MyMultiDirChooser(parent, title="Select Directories", defaultPath=os.getcwd(), agwStyle=MDD.DD_MULTIPLE|MDD.DD_DIR_MUST_EXIST))

Roshgar:
Will get around to putting that together.
But, yeah, i'm not that familiar with object-oriented python. And a bit
rusty with objects so i entirely forgot that the first one was possible. :-/

Roshgar:
I believe we should be good. Re-ran tests, correctly selects and returns the directories, code is DRY and i don't think i missed something else. (But you never know.)

So, i'm going to be using this in a mini-project and probably maintaining/updating it somewhat. I did see the pull that should wrap results in quotes, but i also notice that the return of a multiFileChooser/multiDirChooser is a string with the paths simply separated by spaces. 
I do not notice a single quote except when printing the whole `args.files` and then it's just python's representation of a string in a list.
So i am somewhat curious as to how exactly i get results that handle files with spaces without rewriting over Gooey's install as i currently do.
I also may just be blind again and missing a detail.

chriskiehl:
Merged! I made a few minor tweaks to it (you were passing `*args, **kwargs` to your `super` call without supplying them in the `__init__` method. Other than that, good job! 

Thanks for the feature!

lrq3000:
Great! Thank you very much, I was looking for that feature for Gooey to support my next script, this will ease my life A LOT!

PR #94, opened 2015-07-17
Make checkbox labels clickable

Shura1oplot:
On a native Windows GUI checkbox labels are clickable.
The change makes a checkbox to change its state when the label next to it is clicked.


chriskiehl:
Nice! Thanks!

Issue #58, opened 2015-01-07
PyInstaller executable issue


ardoramor:
After testing the application to work from source, I've built an executable:
   pyinstaller.exe --onefile app.py

When I run the app.exe from the cmd, it prints the following error to the log:

Traceback (most recent call last):
  File "<string>", line 146, in <module>
  File "C:\...\build\app\out00-PYZ.pyz\gooey.python_bindings.gooey_decorator", line 86, in inner
IOError: [Errno 2] No such file or directory: 'C:\\Users\\...\\AppData\\Local\\Temp\\_MEI108682\\gooey\\_tmp\\app.exe'

Does Gooey play well with PyInstaller or it hasn't been tested?

chriskiehl:
You're a bit of a pioneer :) I haven't dug into PyInstaller and what all Gooey needs in order to be packaged up nicely. 

Try running it without --onefile. I know in the past there was an issue with PyInstaller that, for whatever reason, would cause it to throw an error whenever you packaged a python app that used `multiprocessing` with `--onefile`. I dunno if that's still the case, but it could be a good starting point for debugging. 

I'll hopefully dig into the issue soon.

ardoramor:
I got the following output while compiling without --onefile:

Traceback (most recent call last):
  File "<string>", line 151, in <module>
  File "C:\Users\...\Documents\Projects\app\build\App\out00-PYZ.pyz\gooey.python_bindings.gooey_decorator", line 86, in inner
IOError: [Errno 2] No such file or directory: 'C:\\Users\\...\\DOCUME~1\\PROJ~1\\Ap~1\\dist\\Ap~1\\gooey\\_tmp\\App.exe'

The following directory exists:
    'C:\\Users\\...\\DOCUME~1\\PROJ~1\\Ap~1\\dist\\Ap~1\\
But gooey\\_tmp\\App.exe does not exist.

Looking at PyInstaller's warning output, I noticed the following:
   W: delayed  __import__ hack detected at line 0 - gooey.python_bindings.modules (C:\Python27\lib\site-packages\gooey-0.1.0-py2.7.egg\gooey\python_bindings\modules.pyc)
   W: delayed  __import__ hack detected at line 0 - gooey.python_bindings.gooey_decorator (C:\Python27\lib\site-packages\gooey-0.1.0-py2.7.egg\gooey\python_bindings\gooey_decorator.pyc)

thoppe:
[...]

PR #97, opened 2015-08-12
Don't pass frozen programs to python.

8BitAce:
Fixes part of #58 (thanks to @daranguiz for this solution).

If a script has been frozen, Gooey should run the executable as-is instead of trying to pass it to Python. This should work with most if not all Python "freezing" solutions, specifically PyInstaller and newer versions of Py2exe which both have a frozen attribute in sys.


chriskiehl:
Thanks! :+1:

PR #114, opened 2015-10-09
Set default state in a CheckBox

joaquins:
Although using a default value in Argparse store_true/store_false/store_const seems to not make much sense, it does make sense to have it in the widget itself.

This way, when using a pre-generated JSON configuration, the default values are actually working in the CheckBox (toggling it as active if the default value is True)

### DIFF
--- a/gooey/gui/widgets/components.py
+++ b/gooey/gui/widgets/components.py
@@ -112,6 +112,7 @@ class CheckBox(BaseGuiComponent):
     self.widget = None
     #  data
     self.option_strings = data['commands'][0]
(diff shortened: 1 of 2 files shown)


chriskiehl:
Cool beans. Works well! 

Only thing: could you clean up the pull a bit? It has code from #115 (which needs a few changes). Pull that bit out and I'll click this here merge button! 

Thanks!

joaquins:
Ops! Git-rookie mistake, I send the pull request from my master branch and the commits kept going there...

Please close this one, I just sent a proper pull request from a new branch ;)

PR #115, opened 2015-10-09
Option to load previously generated configuration in the Gooey decorator

joaquins:
Option to load previously generated configuration using the Gooey decorator. New load_build_config parameter added.

This makes it easy to dump the configuration, and do some tweaking to use it later, if needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,7 +96,9 @@ Different styling and functionality can be configured by passing arguments into 
            default_size=(610, 530),   # starting size of the GUI
            required_cols=1,           # number of columns in the "Required" section
            optional_cols=2,           # number of columbs in the "Optional" section
-           dump_build_config=False)   # Dump the JSON Gooey uses to configure itself
+           dump_build_config=False,   # Dump the JSON Gooey uses to configure itself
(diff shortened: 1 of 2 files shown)


chriskiehl:
Looks pretty good! Just a couple of things:

1. I think the the exception on line 53 should halt the execution of the program. It took me a few runs to realize that it was displaying the error message in the console due to it failing to parse my config. Because of the UI nature of Gooey, once it's up and running I expect any kind of status messages/errors to be pushed through the UI, rather than pushed to a console that's out of primary focus. So it initially looked like the load feature was broken until I noticed the message in PyCharm's console.  

2. Could you update the README with the new decorator addition?

joaquins:
Done... I also added the monospace_display option to the README. 

BTW: As I'm not confident with reStructuredText, I checked this file at http://rst.ninjs.org/ and it complains of a lot of warnings... please take it a look in case this warnings make sense to be fixed.

chriskiehl:
Good catch on the missing `mono_space` entry. 

The README is in Markdown, not ReST (Thank god!). Changes look good. Merged!

PR #118, opened 2015-10-23
Improve cx_freeze support

Shura1oplot:
Being frozen (e.g. sys.argv[0] has .exe suffix) Gooey app is looking for resources (images, language files) near the executable file.

### DIFF
--- a/gooey/gui/util/quoting.py
+++ b/gooey/gui/util/quoting.py
@@ -1,4 +1,12 @@
-import re
+import sys
+
+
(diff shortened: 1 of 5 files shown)


gooey/gui/util/quoting.py
```diff
@@ -1,4 +1,12 @@
-import re
```
chriskiehl: This import is still needed by `maybe_quote`, which is used by things in `widget_pack.py`

PR #119, opened 2015-10-23
Separate boxes used to input command line parts and other data

Shura1oplot:
Strongly separate input boxes used to data and command line input boxes.
MultiDirChooser and other similar Payloads store not a part of a command but a list of paths separated by a os dependend path separator. Also POSIX shell quoting added but untested.
This pull request contains https://github.com/chriskiehl/Gooey/pull/118

### DIFF
--- a/gooey/gui/widgets/widget_pack.py
+++ b/gooey/gui/widgets/widget_pack.py
@@ -111,20 +111,12 @@ class BaseMultiFileChooser(BaseFileChooser):
       return '{} {}'.format(self.option_string, value)
     return value or ''
 
-  def onButton(self, evt):
-    dlg = self.dialog(self.parent)
-    result = (self.get_path(dlg)
-              if dlg.ShowModal() == wx.ID_OK
-              else None)
-    if result:
-      self.text_box.SetValue(result)
-
   def get_path(self, dlg):
     return os.pathsep.join(dlg.GetPaths())
 
 
 class MyMultiDirChooser(MDD.MultiDirDialog):
-  def __init(self, *args, **kwargs):
+  def __init__(self, *args, **kwargs):
     super(MyMultiDirChooser, self).__init__(*args, **kwargs)
 
   def GetPaths(self):
--- a/gooey/gui/widgets/widget_pack.py
+++ b/gooey/gui/widgets/widget_pack.py
@@ -86,11 +86,10 @@ class BaseFileChooser(BaseChooser):
     self.dialog = dialog
 
   def getValue(self):
-    value = ' '.join(quote(x) for x in self.text_box.GetValue().split(os.pathsep))
+    value = self.text_box.GetValue()
     if self.option_string and value:
-      return '{} {}'.format(self.option_string, value)
-    else:
-      return value or ''
+      return '{} {}'.format(self.option_string, quote(value))
+    return quote(value) if value else ''
 
   def onButton(self, evt):
     dlg = self.dialog(self.parent)
@@ -101,10 +100,30 @@ class BaseFileChooser(BaseChooser):
       self.text_box.SetValue(result)
 
   def get_path(self, dlg):
-    if isinstance(dlg, wx.DirDialog) or isinstance(dlg, CalendarDlg):
-      return dlg.GetPath()
-    else:
-      return os.pathsep.join(dlg.GetPaths())
+    return dlg.GetPath()
+
+
+class BaseMultiFileChooser(BaseFileChooser):
+  def __init__(self, dialog):
+    BaseFileChooser.__init__(self, dialog)
+
+  def getValue(self):
+    value = ' '.join(quote(x) for x in self.text_box.GetValue().split(os.pathsep) if x)
+    if self.option_string and value:
+      return '{} {}'.format(self.option_string, value)
+    return value or ''
+
+  def onButton(self, evt):
+    dlg = self.dialog(self.parent)
+    result = (self.get_path(dlg)
+              if dlg.ShowModal() == wx.ID_OK
+              else None)
+    if result:
+      self.text_box.SetValue(result)
+
+  def get_path(self, dlg):
+    return os.pathsep.join(dlg.GetPaths())
+
 
 class MyMultiDirChooser(MDD.MultiDirDialog):
   def __init(self, *args, **kwargs):
@@ -122,10 +141,10 @@ def build_dialog(style, exist_constraint=True, **kwargs):
 
 FileChooserPayload    = partial(BaseFileChooser, dialog=build_dialog(wx.FD_OPEN))
 FileSaverPayload      = partial(BaseFileChooser, dialog=build_dialog(wx.FD_SAVE, False, defaultFile="Enter Filename"))
-MultiFileSaverPayload = partial(BaseFileChooser, dialog=build_dialog(wx.FD_MULTIPLE, False))
+MultiFileSaverPayload = partial(BaseMultiFileChooser, dialog=build_dialog(wx.FD_MULTIPLE, False))
 DirChooserPayload     = partial(BaseFileChooser, dialog=lambda parent: wx.DirDialog(parent, 'Select Directory', style=wx.DD_DEFAULT_STYLE))
 DateChooserPayload    = partial(BaseFileChooser, dialog=CalendarDlg)
-MultiDirChooserPayload = partial(BaseFileChooser, dialog=lambda parent: MyMultiDirChooser(parent, title="Select Directories", defaultPath=os.getcwd(), agwStyle=MDD.DD_MULTIPLE|MDD.DD_DIR_MUST_EXIST))
+MultiDirChooserPayload = partial(BaseMultiFileChooser, dialog=lambda parent: MyMultiDirChooser(parent, title="Select Directories", defaultPath=os.getcwd(), agwStyle=MDD.DD_MULTIPLE|MDD.DD_DIR_MUST_EXIST))
 
 class TextInputPayload(WidgetPack):
   def __init__(self, no_quoting=False):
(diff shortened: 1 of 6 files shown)


Shura1oplot:
The main idea of the changes is to separate logic of what users see in GUI input boxes and how command line are build. It is more about design than functionality.

For example (code from `widget_demo.py`):
```python
import sys

@Gooey
def main():
  desc = "Example application to show Gooey's various widgets"
  file_help_msg = "Name of the file you want to process"
  my_cool_parser = GooeyParser(description=desc)
  my_cool_parser.add_argument("-w", "--writelog", default="No, NOT whatevs", help="write log to some file or something")
  args = my_cool_parser.parse_args()

if __name__ == '__main__':
  print sys.argv
  main()
```
This code raises an argparse exception and the args are `['widget_demo.py', '-w', 'No,', 'NOT', 'whatevs']`. To avoid this user have to quote text manually. This behaviour seems not to be clear. Another example, to open file `-h` you need to set text in an input box to `-- -h`. Generally, I feel "lack" of abstraction.

I confirm that I've broken some widgets. The things to note you wrote is applicable to BaseFileChooser and Choosers inherited from it. The others have something like `return quote(value) if value else ''` and work correctly.

I didn't considered `nargs` making this changes. Maybe the best way is to use unquoted text from boxes if `nargs` is set to `*` or `+`. I need to think about this.

If you like the idea, I'll rework this pull request to fix Chooser I broke.

chriskiehl:
@Shura1oplot Ah, I see now. Thanks for the example! 

These are a lot of changed files in this pull. Could you give just a bit of context as to what was added/changed? I see new base classes, and some changes to how arguments are collected. Just to make reviewing this a little easier, could you give a quick rundown of the moving parts?

gooey/gui/widgets/widget_pack.py
```diff
@@ -48,9 +52,9 @@ def __init__(self, button_text=''):
     self.text_box = None
     self.button = None
 
-  def build(self, parent, data=None):
+  def build(self, parent, data):
```
Shura1oplot: `data` cannot be `None` as it expected to be a `dict`

gooey/gui/widgets/widget_pack.py
```diff
@@ -212,4 +249,4 @@ def getValue(self):
 
 def safe_default(data, default):
```
Shura1oplot: linter error: unused argument `default`

gooey/gui/windows/advanced_config.py
```diff
@@ -93,10 +93,12 @@ def GetOptions(self):
...
-    values = [c.GetValue()
-              for c in chain(*self.widgets)
-              if c.GetValue() is not None]
-    return ' '.join(values)
+    _f = lambda lst: [x for x in lst if x is not None]
+    optional_args = _f([c.GetValue() for c in self.widgets.optional_args])
+    required_args = _f([c.GetValue() for c in self.widgets.required_args if c.HasOptionString()])
+    position_args = _f([c.GetValue() for c in self.widgets.required_args if not c.HasOptionString()])
+    if position_args: position_args.insert(0, "--")
+    return ' '.join(chain(required_args, optional_args, position_args))
```
Shura1oplot: Change command line arguments order: [program name] [command (subparser)] [required named arguments] [optional named arguments] -- [positional arguments]. Adding `--` helps argparse to separate named arguments and positional. For instance, you can open file `-h`. Also it sometimes is necessary with `nargs`.

gooey/gui/widgets/widget_pack.py
```diff
@@ -37,7 +36,12 @@ def onResize(self, evt):
   def get_command(data):
     return data['commands'][0] if data['commands'] else ''
 
-
+  @staticmethod
+  def disable_quoting(data):
```
Shura1oplot: New quoting policy is not applicable if `nargs` is set to `*`, `+` or any other values, except `1` and `?`.

gooey/gui/widgets/widget_pack.py
```diff
@@ -95,15 +98,34 @@ def onButton(self, evt):
...
   def get_path(self, dlg):
-    if isinstance(dlg, wx.DirDialog) or isinstance(dlg, CalendarDlg):
-      return maybe_quote(dlg.GetPath())
-    else:
-      paths = dlg.GetPaths()
-      return maybe_quote(paths[0]) if len(paths) < 2 else ' '.join(map(maybe_quote, paths))
+    return dlg.GetPath()
+
+
+class BaseMultiFileChooser(BaseFileChooser):
```
Shura1oplot: Сlass `BaseFileChooser` is splitted into two classes. Before that we have `if` condition in `get_path()` and `getValue()` methods to check whether Chooser is used for multi file selection of not. Now class for multi files is splitted and appropriate `*Payload` objects use explicitly defined class.

gooey/gui/widgets/widget_pack.py
```diff
@@ -117,18 +139,21 @@ def build_dialog(style, exist_constraint=True, **kwargs):
 
 FileChooserPayload    = partial(BaseFileChooser, dialog=build_dialog(wx.FD_OPEN))
 FileSaverPayload      = partial(BaseFileChooser, dialog=build_dialog(wx.FD_SAVE, False, defaultFile="Enter Filename"))
-MultiFileSaverPayload = partial(BaseFileChooser, dialog=build_dialog(wx.FD_MULTIPLE, False))
+MultiFileSaverPayload = partial(BaseMultiFileChooser, dialog=build_dialog(wx.FD_MULTIPLE, False))
```
Shura1oplot: Use new class for multi file Choosers

gooey/gui/widgets/widget_pack.py
```diff
@@ -37,7 +36,12 @@ def onResize(self, evt):
   def get_command(data):
     return data['commands'][0] if data['commands'] else ''
 
-
+  @staticmethod
+  def disable_quoting(data):
+    nargs = data.get('nargs', None)
+    if not nargs:
+      return False
+    return nargs not in (1, '?')
```
Shura1oplot: New quoting policy is not applicable if `nargs` is set to `*`, `+` or any other values, except `1` and `?`.

gooey/gui/widgets/widget_pack.py
```diff
@@ -48,9 +52,9 @@ def __init__(self, button_text=''):
     self.text_box = None
     self.button = None
 
-  def build(self, parent, data=None):
+  def build(self, parent, data):
     self.parent = parent
-    self.option_string = data['commands'][0] if data['commands'] else ''
+    self.option_string = self.get_command(data)
```
Shura1oplot: `data` cannot be `None` as it is expected to be a `dict`

gooey/gui/widgets/widget_pack.py
```diff
@@ -95,15 +98,26 @@ def onButton(self, evt):
...
+    BaseFileChooser.__init__(self, dialog)
+
+  def getValue(self):
+    value = ' '.join(quote(x) for x in self.text_box.GetValue().split(os.pathsep) if x)
+    if self.option_string and value:
+      return '{} {}'.format(self.option_string, value)
+    return value or ''
+
+  def get_path(self, dlg):
+    return os.pathsep.join(dlg.GetPaths())
```
Shura1oplot: Сlass `BaseFileChooser` is splitted into two classes. Before that we have `if` condition in `get_path()` and `getValue()` methods to check whether Chooser is used for multi file selection of not. Now class for multi files is splitted and appropriate `*Payload` objects use explicitly defined class.

Shura1oplot:
@chriskiehl I develop some small tools to work with csv files, so my users have to use `"` and ` ` (space) as command line agruments. That is why I make this changes. I commented some changes in the code. Also please see this example:
```python
import sys

@Gooey
def main():
  parser = GooeyParser()

  parser.add_argument("--quoting", default='"')
  parser.add_argument("--space", default=" ")

  parser.add_argument("--optional_numbers", default="1 2 3", nargs="+")
  parser.add_argument("required_numbers", default="4 5 6", nargs="+")
  args = parser.parse_args()
  print sys.argv

if __name__ == '__main__':
  main()
```
Note that `--optional_numbers` and `required_numbers` are not mixed because of `--` insertion between them.

gooey/gui/widgets/widget_pack.py
bilderbuchi: Empty dict would evaluate to `False`, though.

Shura1oplot:
```>>> data = None
>>> data['commands'][0]
  File "<stdin>", line 1, in <module>
TypeError: 'NoneType' object has no attribute '__getitem__'```

Comment (bilderbuchi) [inline gooey/gui/widgets/widget_pack.py]: this way works, though: 
```
>>> data = None
>>> data['commands'][0] if data else ''
''
```
be aware of this, though (so as it was before won't work):
```
>>> data['commands'][0] if data['commands'] else ''
  File "<stdin>", line 1, in <module>
TypeError: 'NoneType' object is not subscriptable
```

**Comment by chriskiehl** [comment]
Thanks for the additional details! 

:+1: good call on the `--` addition. Smart! 

My only remaining comment would be that [this block](https://github.com/chriskiehl/Gooey/pull/119#discussion-diff-44222244R101) could be simplified a bit. Your `_f` functionality exists as a python builtin via `filter(None, [])`. A potentially less busy refactor could be:

```python
optional_args = [c.GetValue() for c in self.widgets.optional_args]
required_args = [c.GetValue() for c in self.widgets.required_args if c.HasOptionString()]
position_args = [c.GetValue() for c in self.widgets.required_args if not c.HasOptionString()]
if position_args: position_args.insert(0, "--")
return ' '.join(filter(None, chain(required_args, optional_args, position_args)))
```

But that's just me being nit-picky! Rest looks good. Merging in!

Shura1oplot:
@chriskiehl Thanks! You are right about `filter(None, [])` in this case. Now I don't remmember why I checked values for exactly `None` instead of truthfulness as there is no single `GetValue` returning `None` or any other value different from string.

In your variant `--` can be inserted into list full of empty strings as you apply `filter` after insertion.

PR #120, opened 2015-10-23
Implement Stop button and progress bar

Shura1oplot:
Added Stop button to the screen with status of running program. Depend on the OS it uses os.kill to terminale the process or TASKKILL.EXE windows program.

Added progress bar support. Two new `Gooey()` options added: `progress_regex` and `progress_expr`. The first one should be regular expression which is applied to each line of the program output. At least one group in the regex should be defined. If the regexp matchs a line, progress bar is updated. In case `progress_expr` is not set, the group should contain a float number between 0 and 100, otherwise `progress_expr` is applied as an argument to `expr()` function. All unnamed regexp groups are passed as `x` variable with `list` type, named groups are passed as variables. If a group value can be converted to float, it will be. For usage please refer to `example_progress_bar_*.py`.

gooey_progress_bar


chriskiehl:
Hey @Shura1oplot!

This is awesome stuff! I'll start reviewing all these pull request shortly... Just a little buried under other projects right now, but I don't want you to think I'm ignoring these : )

Shura1oplot:
@chriskiehl I reverted "make quoting implicit" commit as we are discussing it in the different PR.

gooey/gui/controller.py
```diff
@@ -108,8 +171,8 @@ def required_section_complete(self):
   def success_dialog(self):
     self.show_dialog(i18n._("execution_finished"), i18n._('success_message'), wx.ICON_INFORMATION)
 
-  def error_dialog(self, error_msg):
-    self.show_dialog(i18n._('error_title'), i18n._('uh_oh').format(error_msg), wx.ICON_ERROR)
```
chriskiehl: Why did you drop the error here (and in the translations file)? 

Gooey should bubble up _why_ the program failed.

gooey/gui/controller.py
```diff
@@ -85,16 +114,50 @@ def read_stdout(self, process, callback):
       if not line:
         break
       wx.CallAfter(self.core_gui.PublishConsoleMsg, line)
+      progress = self.progress_from_line(line)
+      if progress is not None:
+        wx.CallAfter(self.core_gui.UpdateProgressBar, progress)
     wx.CallAfter(callback, process)
 
+  def progress_from_line(self, text):
+    progress_regex = self.build_spec['progress_regex']
```
chriskiehl: It's a little tough to tell what the method is actually doing due to all the error handling boilerplate. I think if you pulled out some of the functionality into methods, you could expose the code's intent a little more clearly. For instance, factoring out the evaluation and conversion bits into their own handlers could give something like this:

```python

  def progress_from_line(self, text):
    progress_regex = self.build_spec['progress_regex'] or ''
    match = re.search(progress_regex, text.strip())
    if not match.group():
      return None
    progress_expr = self.build_spec['progress_expr']
    value = self.evaluate(progress_expr, match) if progress_expr else match.group(1)
    return safe_int(value)
```

gooey/python_bindings/gooey_decorator.py
```diff
@@ -33,7 +33,10 @@ def Gooey(f=None,
           optional_cols=2,
           dump_build_config=False,
           load_build_config=None,
-          monospace_display=False):
+          monospace_display=False,
+          progress_regex=None,
+          progress_expr=None,
+          progress_animation=True):
```
chriskiehl: The display of the `stop` button should be configurable here.

gooey/_tmp/example_progress_bar_4.py
```diff
@@ -0,0 +1,39 @@
...
+
+from __future__ import unicode_literals
+from __future__ import print_function
+import sys
+from time import sleep
+from gooey import Gooey, GooeyParser
+
+
+@Gooey(progress_regex=r"^progress: (-?\d+)%$",
+       progress_animation=False)
```
chriskiehl: Hey, could you expand on this  `progress_animation`  a bit? I see that it's used in `base_window.py` with a comment about Windows 7, but I'm not totally clear on why it's in the decorator options.

gooey/_tmp/example_progress_bar_3.py
```diff
@@ -0,0 +1,26 @@
...
+
+from __future__ import unicode_literals
+from __future__ import print_function
+import sys
+from time import sleep
+from gooey import Gooey, GooeyParser
+
+
+@Gooey(progress_regex=r"^progress: (?P<current>\d+)/(?P<total>\d+)$",
+       progress_expr="current / total * 100")
```
chriskiehl: This is really cool! :+1: :)

gooey/gui/controller.py
Shura1oplot: @chriskiehl Gooey catches stderr to show python exceptions in a error dialog. It doesn't work as designed when wrapped program uses stderr to print messages to user and stdout to optional data out. Lots of unix tools work like this, for example, wget and dd. Also it solves buffered output from frozen programs as stdout is always unbuffered.

gooey/gui/controller.py
Shura1oplot: Ok, I'll refactor this.

gooey/_tmp/example_progress_bar_4.py
Shura1oplot: @chriskiehl Please see [this](http://stackoverflow.com/questions/5332616/disabling-net-progressbar-animation-when-changing-value). Windows 7 progress bar animation sometimes too slow and lagging from real progress if it changes fast. I've got a log for 20% of the progress.

gooey/python_bindings/gooey_decorator.py
Shura1oplot: @chriskiehl Would you like just to hide stop button or hide it and move progress bar to the right of the footer (as it was before)?

chriskiehl: Leave it where it is. I like where you put it. It should just have the option of being hidden. 

There are certainly a subset of programs that shouldn't be casually stopped mid-execution by a user. My personal use-case for Gooey has always been for targeting non-techies. I don't want them to have `taskill` at their fingertips anytime they get impatient : )

gooey/gui/controller.py
chriskiehl: @Shura1oplot Hmm... this is interesting. I didn't know it was common to pump non error messages over stderr. You're exposing my noobish level of linux knowledge! ; )  

After spending a few minutes reading up on the behavior, I yield to your explanation. I can see why programs would use stderr for things other than a simple error stream. 

So... as you said, it seems Gooey's original behavior was indeed broken. 

Good fix!

Shura1oplot:
@chriskiehl excuse me for a long delay, I have "busy season" on my work. I have fixed thing noted above.

chriskiehl:
Merged! Awesome work, man!

PR #128, opened 2015-11-30
Make tests pass, close #127.

bilderbuchi:
build_radio_group() started returning results not wrapped
in a list in 8103314, change the test to reflect that.

PR #129, opened 2015-11-30
More tests

bilderbuchi:
I picked a couple of low-hanging fruit to increase test coverage.
I also found an apparently broken function and removed it, see cc2c095.
This also includes PR #128, as I wanted to start with all tests passing, I can rebase if you want.

I duplicated a fixture from argparse_to_json_unittest, see the TODO marker. This should probably go into a new conftest.py, so that tests can share one implementation. If that's alright with you I can add one.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,8 +23,10 @@ pip-log.txt
 
 # Unit test / coverage reports
 .coverage
+htmlcov
 .tox
 nosetests.xml
+.cache
 
 # Translations
 *.mo
--- a/gooey/python_bindings/config_generator.py
+++ b/gooey/python_bindings/config_generator.py
@@ -41,8 +41,3 @@ def create_from_parser(parser, source_path, **kwargs):
     build_spec['manual_start'] = True
 
   return build_spec
-
-
-
-def has_argparse(module_path):
-  return any(['.parse_args(' in line.lower() for line in f.readlines()])
--- a/gooey/tests/argparse_to_json_unittest.py
+++ b/gooey/tests/argparse_to_json_unittest.py
@@ -2,66 +2,6 @@ import pytest
 from gooey.python_bindings.argparse_to_json import *
 
 
-@pytest.fixture
-def empty_parser():
-  return argparse.ArgumentParser(description='description')
-
-@pytest.fixture
-def complete_parser():
-  parser = argparse.ArgumentParser(description='description')
-  parser.add_argument("req1", help='filename help msg')  # positional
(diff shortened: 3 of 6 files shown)


chriskiehl:
My test coverage is pretty horrendous, so add as much as you want! : ) 

I'm in favor of a `conftest.py` rather than introducing duplication.

bilderbuchi:
OK, I moved all the fixture definitions into a conftest.py. You can review the changes now.

bilderbuchi:
Two remarks: 
* I found out that the codebase basically does not adhere to PEP8 at all, which makes my linters light up like a christmas tree :christmas_tree:. Is this by design/preference? Would you like to see that fixed? Fixing PEP8 compliance would mean touching all the lines (due to indentation being multiple of 2, not 4), so I'm hesitant to add those noisy changes before asking about it.
* I found out that all the files in the repo are with CRLF line endings (probably you're on windows), but they keep those line endings when checked out on my Linux machine, where new files default to LF endings, so that makes some mess. I'm still working on how to best fix this or make it convert line endings on checkout/commit.

chriskiehl:
I tend to ignore linters while working on Gooey ;) The main reason is that WxPython, which makes up _most_ of what Gooey is, uses the style of its parent WX project, which is very much _not_ PEP8 compliant. In the beginning, for the sake of a consistent style across the project, I tried to honor Wx style rather than PEP8. However, this ended up being just as weird as trying to PEP8 outright (for instance, having wxPython's CamelCase and the Python stdlib's snake_case all conflicting on the same line). 

Now, that's not to say that there aren't things that could certainly be fixed. However, I lean towards the lazy "eh, why bother" side of the spectrum. Even if you PEP8 all non-wx code, the boundary line between WX will still be very weird and non-standard. 

The only way to _really_ PEP8 the core project would be to write adapater layers ontop of Wx that PEP8-izes its external interface. _Then_ we could have a gloriously compliant codebase with a tiny adapater section of ugly wx core code, _but..._ the violations don't bother me enough to actually do all that work just for a pep8 checkmark :) 

_All that said__, the CRLF thing is indeed annoying. I thought I had git set up to convert them. I'll clean those up lickity split!

bilderbuchi:
Thanks for the merge!

WyPython style: I see now where the style is coming from, thanks for clearing that up!

> All that said_, the CRLF thing is indeed annoying. I thought I had git set up to convert them. I'll clean those up lickity split! 

I've already done that work for you, see #131. ;-)

PR #137, opened 2015-12-25
Fix progress_bar merging

Shura1oplot:
PR with progress bar implementation merged not fully.

### DIFF
--- a/gooey/gui/util/quoting.py
+++ b/gooey/gui/util/quoting.py
@@ -1,6 +1,4 @@
-import re
 import sys
-import re
 
 
 if sys.platform.startswith("win"):
@@ -9,7 +7,3 @@ if sys.platform.startswith("win"):
 else:  # POSIX shell
   def quote(value):
     return "'{}'".format('{}'.format(value).replace("'", "'\\''"))
-
-
-def maybe_quote(string):
-  return '"{}"'.format(string) if not re.match(r'^".*"$', string) else string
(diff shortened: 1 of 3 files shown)


gooey/gui/util/quoting.py
```diff
@@ -9,7 +7,3 @@ def quote(value):
 else:  # POSIX shell
   def quote(value):
     return "'{}'".format('{}'.format(value).replace("'", "'\\''"))
-
-
-def maybe_quote(string):
```
Shura1oplot: it is useless now

gooey/gui/windows/layouts.py
```diff
@@ -11,7 +11,7 @@
 
 basic_config = {
     'required': [{
-      'type': 'TextField',
+      'type': 'CommandField',
```
Shura1oplot: with new aggressive quoting we shoukd use text field without quoting

gooey/gui/widgets/components.py
```diff
@@ -243,6 +243,7 @@ def _GetWidget(self):
 FileSaver         = lambda data: BaseGuiComponent(data=data, widget_pack=widget_pack.FileSaverPayload())
 DateChooser       = lambda data: BaseGuiComponent(data=data, widget_pack=widget_pack.DateChooserPayload())
 TextField         = lambda data: BaseGuiComponent(data=data, widget_pack=widget_pack.TextInputPayload())
+CommandField      = lambda data: BaseGuiComponent(data=data, widget_pack=widget_pack.TextInputPayload(no_qouting=True))
```
chriskiehl: `no_qouting` -> `no_quoting` :)

PR #138, opened 2016-01-04
Improve stop button

Shura1oplot:
Suggesting improving stop button:
* no more blinking console window if program was frozen as win32gui app
* try to stop program gently on Linux and OSX: first pressing to stop button sends SIGINT to the program, second - SIGTERM, third and further - SIGKILL. Close button (cross) always sends SIGKILL. No any idea how to do this on Windows.

Example below shows the program that stops working only by third pressing of stop button (or immediately - on cross).
```python
#!/usr/bin/env python
# -*- coding: utf-8 -*-

from __future__ import unicode_literals
from __future__ import print_function

import sys
from time import sleep


import signal
signal.signal(signal.SIGINT, lambda *_: None)  # ignore first attempt to stop
signal.signal(signal.SIGTERM, lambda *_: None)  # ignore second attempt to stop
# it is imposible to handle or ignore third attempt to stop - SIGKILL


@Gooey(progress_regex=r"^progress: (\d+)%$")
def main():
    parser = GooeyParser(prog="example_progress_bar_1")
    _ = parser.parse_args(sys.argv[1:])

    for i in range(100):
        print("progress: {}%".format(i+1))
        sys.stdout.flush()
        sleep(0.1)


if __name__ == "__main__":
    sys.exit(main())
```

### DIFF
--- a/gooey/gui/controller.py
+++ b/gooey/gui/controller.py
@@ -38,7 +38,7 @@ class Controller(object):
     self.core_gui = base_frame
     self.build_spec = build_spec
     self._process = None
-    self._task_terminated = False
+    self._stop_pressed_times = 0
 
     # wire up all the observers
     pub.subscribe(self.on_cancel,   events.WINDOW_CANCEL)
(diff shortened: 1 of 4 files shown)


chriskiehl:
Hey @Shura1oplot, 

You'll have to revisit this. I changed a buncha code underneath you : \  (sorry!) 

The bulk of your progressbar code now lives in `progressor.py`. It in turn publishes its updates to the view via the Presentation Model in `presentation.py`

Shura1oplot:
@chriskiehl, I have missed it :( Well, I need some time to update PR.

PR #154, opened 2016-02-24
Improve Spanish translation

jslirola:
Translate the last string and to do minor fixes.


chriskiehl:
Cool Beans. Thanks :+1:

PR #160, opened 2016-06-15
Fix subprocess to work properly when frozen

bastula:
This fixes frozen Gooey scripts when frozen using PyInstaller. #58 
See: http://stackoverflow.com/a/10338737/74123


chriskiehl:
Shazzam! A year later this gets a ticket to merge town! 

Thanks for the patch!

bastula:
Gosh I even forgot that I made this PR 🤣. Thanks for merging it in!

PR #163, opened 2016-07-05
fix crash when using subparsers

sodwyer:
switching view to a different subparser group crashed on this line because populate expects 2 arguments


chriskiehl:
:+1:

PR #174, opened 2016-11-26
Update README.md

jjconti:
Use the correct widget name.


chriskiehl:
#173 beat ya with the same fix. : )

PR #180, opened 2017-01-14
Add 'parse_only' option to receive arguments from GUI and return to CLI

crowjdh:
Some terminal-specific api calls(ex, curse.cbreak()) yields error and can't be adopted to current Gooey(wxPython) environment.
Receiving arguments from GUI and returning them to original process(mostly CLI) may resolve this issue, by supplying "Settings" window only and calling callback function when "Start" button is clicked.

I couldn't come up with any better name for the option "parse_only", though it may not be intuitive.

### DIFF
--- a/README.md
+++ b/README.md
@@ -247,7 +247,7 @@ Just about everything in Gooey can be customized by passing arguments to the dec
 | advanced | Toggles whether to show the 'full' configuration screen, or a simplified version | 
 | show_config | Skips the configuration all together and runs the program immediately |
 | language | Tells Gooey which language set to load from the `gooey/languages` directory.|
-|program_name | The name displayed in the title bar of the GUI window. If not supplied, the title defaults to the script name pulled from `sys.argv[0]`. |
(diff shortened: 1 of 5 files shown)


lrq3000:
This is a wonderful idea, it would fix the compatibility with most console applications while still providing a nice interface for the user to provide the arguments!

chriskiehl:
I think this is a neat idea. Sorry it took so long to get around to commenting on it! 

Work moved forward, and most of the code you modified is now completely removed from the project due to a much needed project cleanup pass. If you're interested in updating the code, I'd be happy to merge it! 

I think the main thing that changed from the time that you created this pull is that `models.py` and `presenter.py` (two deeply embarrassing files created by past me) have been removed. In their place is now just a plain ol' fat controller/view file under `/gooey/gui/containers/application.py`. And the building of CLI args is now handled via a `gooey/gui/cli.py`.

crowjdh:
Closing due to newer PR #241

PR #182, opened 2017-01-19
Greek Translation

MikeMeliz:
Good morning everybody,
That's my Greek translation for this awesome tool!
:)


chriskiehl:
Thanks for the language pack! Merged!

PR #202, opened 2017-06-24
Adding ability to override executable target

lzulauf:
We use a custom executable to configure our python environment, but the logic for determining the target assumes we are using the sys.program or are frozen.

Adding the ability to specify a target override.


chriskiehl:
Code lgtm. However, could you update the Usage section of the README with your new argument? After that: merge town!

lzulauf:
Updated README. let me know if this looks good

chriskiehl:
👍

chriskiehl:
Pulled. Thanks!

Issue #200, opened 2017-06-09
Feature request: Substitute command-line arguments for Gooey defaults


jschultz:
I've been trying to work out a simple way to combine command-line and GUI, and it struck me that it would be nice if arguments specified on the command-line could automagically appear as default values in the resulting GUI.

I tried reverse engineering, (and found the very useful but apparently undocumented '--ignore-gooey' option!) but it seems I'm a long way down the learning curve and thought it better to ask the experts who may be able to implement this in 10 minutes, explain why it is not a good idea, or else offer some pointers and I'll work on it myself.

What do people think?

jjconti:
Thanks for this! `--ignore-gooey`

zertrin:
Actually this is something that, as a first time user, I was kind of expecting to work already. I strongly support this feature request!

chriskiehl:
Merged into release.

PR #203, opened 2017-06-26
Features to allow easier use of command line with Gooey

jschultz:
This pull request incorporates three related changes.

1. The simplest is really a bug fix. Previously if you used the `@Gooey` decorator on a function with one or more arguments and called the parser with `--ignore-gooey`, Python failed with

> TypeError: <lambda>() takes no arguments (1 given)

This was fixed by modifying the `run_without_gooey` function to used an `inner2` function like `run_gooey` does.

2. The second modification allows the IGNORE_COMMAND argument to be specified as an argument (`ignore_command`)  to the `@Gooey` decorator (I did this because prefer `--no-gui` to  `--ignore-gooey`) and adds a further argument (`force_command`) to be specified. The latter argument reverses the default functionality, so that the script is run without Gooey unless the argument is specified. For example:

> my_command --gui

3. A slightly more complicated change responds to #200 by adding a boolean argument `load_cmd_args` to the `@Gooey` decorator.  If this argument is specified then Gooey will parse the arguments from the command line invocation, and use the resulting values as default values (over-riding defaults specified in the Python script) when the GUI is generated. Note that this functionality could be improved by allowing positional arguments to be skipped and not enforcing `required` when parsing the command line arguments.

### DIFF
--- a/gooey/gui/model.py
+++ b/gooey/gui/model.py
@@ -219,7 +219,7 @@ class MyModel(object):
     cmd_string = ' '.join(list(filter(None, chain(required_args, optional_args, position_args))))
     if self.layout_type == 'column':
       cmd_string = u'{} {}'.format(self.argument_groups[self.active_group].command, cmd_string)
-    return u'{} --ignore-gooey {}'.format(self.build_spec['target'], cmd_string)
+    return u'{} {} {}'.format(self.build_spec['target'], self.build_spec['ignore_command'] or '', cmd_string)
 
   def group_arguments(self, widget_list):
     is_required = lambda widget: widget['required']
--- a/gooey/python_bindings/argparse_to_json.py
+++ b/gooey/python_bindings/argparse_to_json.py
@@ -52,7 +52,7 @@ class UnsupportedConfiguration(Exception):
 }
 
 
-def convert(parser):
+def convert(parser, cmd_args=None):
   widget_dict = getattr(parser, 'widgets', {})
   actions = parser._actions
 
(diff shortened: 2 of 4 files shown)


chriskiehl:
I'm not totally sure I understand the use case for your third feature. Gooey has generally been about end users _not_ using the command line. You can specify default args for fields in the `buildspec`. Is the extra path just for testing or dev convenience?

jschultz:
The logic of using the command line is to make it easy to have dynamic default values. I use it in production code to allow last-used values to appear as defaults.

3dp:
Is there some documenation about how to use `--ignore-gooey`, `load_cmd_args`, etc?

jschultz:
No documentation that I know of. `load_cmd_args` simply parses command-line arguments and uses them as defaults for the gui. Could probably be done in user space but nicer to make it part of gooey. Any chance I could convince to you include it @chriskiehl ?

jornb:
I think these changes are great @jschultz. I just started playing around with Gooey, and I instantly wanted both 2 and 3 for my use cases.

zpgaal:
What is the state of the 3rd option,
When will it be part of gooey?
It is really useful. Ex: to check command line argument before pressing start or to adjust them.

dsmic:
I don't know, if you are still working on this, but I would love to have an option to only show the config gui, and than leaving the output unchanged on the console from which I started the script.

Thanks a lot for your fork!

jschultz:
I haven't looked at this lately, but glad to hear it's been useful for you @dsmic. The fork could do with being rebased onto a more current upstream release to get rid of those conflicts.

What I really need is another client who refuses to use command-line! That was the initial inspiration...

Regarding your request, I'm a little unsure what you mean by 'leaving the output unchanged' - which output exactly do you mean and how would it otherwise be changed. Maybe an example would help?

travipross:
Just like to weigh in and say item 2 above would be useful for me as well. I utilize a number of scripts with command line arguments that are executed via batch scripts with no user input. Occasionally, however, these same scripts are utilized with varying arguments by contractors who are less familiar with the command line. Being able to specify "--gui" to trigger a Gooey window, but otherwise fall back on the standard argparse workflow would be a great help.

PR #210, opened 2017-07-10
Set background colour on required field validation

jschultz:
This pull request highlights empty required fields on validation by setting their background colour to red.

### DIFF
--- a/gooey/gui/model.py
+++ b/gooey/gui/model.py
@@ -4,6 +4,7 @@ from itertools import chain
 from gooey.gui.lang.i18n import _
 from gooey.gui.util.quoting import quote
 
+import wx
 
 ArgumentGroup = namedtuple('ArgumentGroup', 'name command required_args optional_args')
 
@@ -207,8 +208,20 @@ class MyModel(object):
     return self.build_spec['manual_start']
 
   def is_required_section_complete(self):
-    completed_values = list(filter(None, [arg.value for arg in self.required_args]))
-    return len(self.required_args) == len(completed_values)
+    error_found = False
+    for arg in self.required_args:
+      widget = arg.widget_instance.widget_pack.widget
+      if arg.value:
+        widget.SetBackgroundColour(wx.NullColour)
+      else:
+        if not error_found:
+          error_found = True
+          widget.SetFocus()
+        widget.SetBackgroundColour("#EF9A9A")
+
+      widget.Refresh()
+
+    return not error_found
 
   def build_command_line_string(self):
     optional_args = [arg.value for arg in self.optional_args]
--- a/gooey/gui/windows/advanced_config.py
+++ b/gooey/gui/windows/advanced_config.py
@@ -51,6 +51,7 @@ class WidgetContainer(wx.Panel):
     for index, widget in enumerate(widgets):
       widget_class = getattr(components, widget.type)
       widget_instance = widget_class(self, widget.title, widget.help, widget.choices)
+      widget.widget_instance = widget_instance
       self.widgets.append(widget_instance)
     self.layout(num_columns)
 


gooey/gui/model.py
```diff
@@ -207,8 +208,18 @@ def skipping_config(self):
...
+    error_found = False
+    for arg in self.required_args:
+      widget = arg.widget_instance.widget_pack.widget
+      if arg.value:
+        widget.SetBackgroundColour(wx.NullColour)
+      else:
+        if not error_found:
+          error_found = True
+          widget.SetFocus()
+        widget.SetBackgroundColour("Red")
```
chriskiehl: I think straight red is a little too bold for Gooey's muted palette. Additionally, you keep the font color the same, so it all end up looking a little wonky. 

Could you soften it up a bit? `#EF9A9A` looked pretty good to my eye.

gooey/gui/model.py
```diff
@@ -207,8 +208,18 @@ def skipping_config(self):
     return self.build_spec['manual_start']
 
   def is_required_section_complete(self):
-    completed_values = filter(None, [arg.value for arg in self.required_args])
-    return len(self.required_args) == len(completed_values)
+    error_found = False
+    for arg in self.required_args:
+      widget = arg.widget_instance.widget_pack.widget
```
chriskiehl: You should call `widget.Refresh()` on everything. When there are multiple required fields, I'm  only seeing the first one change colors until I mouse over the rest.

gooey/gui/model.py
```diff
@@ -207,8 +208,18 @@ def skipping_config(self):
     return self.build_spec['manual_start']
 
   def is_required_section_complete(self):
-    completed_values = filter(None, [arg.value for arg in self.required_args])
```
chriskiehl: Technically... _technically..._ you shouldn't be making UI changes here in the model and instead use the Presenter class, but, due to my many, many code organization sins, you'd probably have to change a whole lot of stuff in order to make that work. Further, all this code all changes in 2.0, so ultimately it fine living here

gooey/gui/model.py
jschultz: Done

gooey/gui/model.py
jschultz: I can't reproduce this. But I've done as you requested, if you can check that it works as intended.

jschultz:
I'm afraid my git skills leave something to be desired. I tried following instructions to rebase, and now it tells me that I'm trying to merge 27 commits?!

bilderbuchi:
What are you trying to accomplish and what did you try?

chriskiehl:
No pain quite like git pain. : ) 

The good news is that git works _really_ hard to not lose your work. Without knowing exactly how you got to the current state, I'd wager that the safest bet is approach this in three steps. 1. do a little reflog spelunking and branch off of your last known good commit to get yourself back in a good state,  (2) clean up your history, (3) bring yourself up to date with master.  

**Recovering your original state:**

It'll look something like this: 

`git reflog`

That'll show you a chronological list of all your git activity. Along the lines of: 

```
a26df8f HEAD@{0}: checkout: Set background colour on required field validation
...
```

You should be able to find your last known good commit before you started rebasing (probably easiest to identify by commit message rather than sha). Once you found that, all you got to do is check it out 

```
git checkout [YOUR_SHA]
```

Then, create a new branch from that point 

```
git checkout -b new-branch-name 
```

Once you done that, a `git log` should show you all the history you expect. 

**2. Fix your local history:**

The easiest way to do this is to not rebase at all, but instead `reset --soft` back to when you started your work. 

```git log --oneline```

it'll print out a list of commits

```
a26df8f you commit 1
bebc05b your commit 2
a7a6421 your commit 3
...
1b2c2e1 a commit from master
```
You'll want to reset soft back to that **first commit that isn't yours**

```
git reset --soft SHA_OF_THE_FIRST_COMMIT_THAT_ISNT_YOURS
```

That'll undo all your commits and put everything back into git staging, which you can verify with 

```
git status
```

You should see all your changes staged. I'd do a `git diff` at this point to verify!

If all looks correct, you can just commit as usual with a nice message. 

```
git commit -m "your feature" 
```


**3. Bringing yourself up to date with master:**

Checkout master and pull

```
git checkout master
git pull
``` 

Then hop back to your branch 

```
git checkout YOUR_BRANCH_NAME
```

And finally, rebase _against_ master 

```
git rebase master
```

Something more or less like that ; )

jschultz:
Thanks for your patience! After messing it up a few more times I finally made it work. Let's see how I go with those other pull requests now.

chriskiehl:
👍

PR #211, opened 2017-07-10
Groups

jschultz:
This pull request adds functionality to group arguments according to argparse groups, rather than the fixed groups `required arguments` and `optional arguments`. With this change, I find that I can now make pretty and usable GUIs for Python scripts with many arguments.

Specifying the argument `use_argparse_groups` to the Gooey wrapper invokes this new functionality - otherwise the operation should remain completely unchanged. Related arguments are `default_cols` which specifies the default number of columns, and `cols_dict` which allows the number of columns to be specified per group.

Note that to work properly, this pull request needs to be  merged with #210, since it is no longer immediately obvious which arguments are required and which are optional, and #209, since the previous version didn't handle optional positional arguments at all.

### DIFF
--- a/gooey/gui/presenter.py
+++ b/gooey/gui/presenter.py
@@ -33,10 +33,12 @@ class Presenter(object):
 
     pub.subscribe(self.on_selection_change, events.LIST_BOX)
 
-
   def on_selection_change(self, selection):
-    self.update_model()
+    self.delete_model()
     self.model.active_group = selection
+    self.create_model()
+    self.view.do_layout()
+    self.update_model()
     self.redraw_from_model()
     self.syncronize_from_model()
 
@@ -44,10 +46,8 @@ class Presenter(object):
     self.view.window_title = self.model.program_name
     self.view.window_size = self.model.default_size
 
-    self.view.required_section.clear()
-    self.view.optional_section.clear()
-    self.view.required_section.populate(self.model.required_args, self.model.num_required_cols)
-    self.view.optional_section.populate(self.model.optional_args, self.model.num_optional_cols)
+    self.create_model()
+    self.view.do_layout()
 
     if self.model.use_monospace_font:
       self.view.set_display_font_style('monospace')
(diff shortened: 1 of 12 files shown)


chriskiehl:
Could you cleanup your commit history into a commit (or two) that describes the addition/change. They're currently a bit chatty and stream of conscious.

gooey/gui/presenter.py
```diff
@@ -65,9 +65,23 @@ def initialize_view(self):
       self.on_start()
     self.syncronize_from_model()
 
+  def create_model(self):
+    for group in self.model.groups():
+      if self.model.args(group):
+        self.view.create_section(group)
+        self.view.section(group).clear()
+        self.view.section(group).populate(self.model.args(group), self.model.num_cols_dict.get(group, self.model.num_default_cols))
```
chriskiehl: Whoa, long line is long! : )

chriskiehl:
image

This line that sits just below the main panel is a little wonky looking.

chriskiehl:
image

image


I'm seeing some layout issues when: 

1. Clicking the start button with no fields filled out
2. Clicking between sections in the subparser view

gooey/gui/presenter.py
jschultz: Fair enough. (I use an editor that does smart wrap, so tend not to notice these things.)

jschultz:
I'm not seeing that wonkiness. Can you tell me your build environment so I can replicate?

chriskiehl:
Above snapshots were via: 

* python 2.7.6
* Windows 10 
* Wx 3.0.2.0 msw (classic)

Interestingly, I don't see the issue on Ubuntu(14.4) which has an older wxpython version (`'2.8.12.1 (gtk2-unicode)`). However, it has its own quirk: tabs render poorly. 

image

jschultz:
To be honest, that looks like a bug in wxPython to me. And I found another bug using wxPython 3.0.2 on Mac where the pane under the first tab wasn't rendered at all, but if you go to the second tab then back to the first everything is fine.

But all this seems to depend on which version of wxPython you are using. Pip now seems to install version 4.0.x (or should I say tries to - I get compilation errors on by Linux box :( ). But Gooey seems to be incompatible with wxPython 4.0, (eg in footer.py you need to import wx.adv not wx.animate, there are problems calling wx.Font that I haven't gotten to the bottom of, and probably more issues down the track.

So, to cut the chase, do you have a plan for moving to more recent versions of wxPython? I'd be happy to help when I get some time, but maybe you already ahead of me on this?

jschultz:
I had another go at this, now that wxPython Phoenix seems stable. Looks OK to me with wxPython 4.0.0b2. Let me know what you think. Cheers!

chriskiehl:
I decided to go a different direction with Groups. It was rolled into the 1.0 cleanup work here: https://github.com/chriskiehl/Gooey/commit/4c61c29b650b2d43ca2476bf62eefdeb5dceb4ef#diff-c2a9e315e78b26f3c7b03697a9f06437

Sorry : /

PR #214, opened 2017-07-17
Kill child processes as well as shell process

jschultz:
Stopping a running script fails under Linux because only the shell process is stopped. This pull request makes the `Stop` button also kill all its child processes, notably the Python process itself, but also any children it may have created.


chriskiehl:
👍

Issue #230, opened 2017-12-09
 Non-ASCII output can't show in display panel on windows system with chinese language


xavierskip:
I use windows 10 and  is a chinese user.
I found print any Non-ASCII character can't show in display panel.
python version: 3.6
this is my sample code:
```
#!/usr/bin/env python
# -*- coding: utf-8 -*-
import sys
import argparse
from gooey import Gooey

@Gooey()
def main():
    parser = argparse.ArgumentParser()
    parser.add_argument('--name')
    args = parser.parse_args()
    print("hello!{}".format(args.name))

if __name__ == "__main__":
    sys.exit(main())
```
when i run it and type in some chinese and execute,I will see warning in terminal like this:
```
Traceback (most recent call last):
File "C:\Program Files (x86)\Python36-32\lib\site-packages\wx\core.py", line 3163, in
<lambda> 
    lambda event: event.callable(*event.args, **event.kw) )
File "C:\Program Files (x86)\Python36-32\lib\site-packages\gooey\gui\windows\runtime_display_panel.py", line 53, in append_text
    self.cmd_textbox.AppendText(txt)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xc4 in position 6: invalid continuation byte
```
I think this is because windows Character Encoding.
```
>chcp
活动代码页: 936
```
I try the same in MacOS and everything works fine.

End I take a patch in`gooey\gui\windows\runtime_display_panel.py` line 53
-self.cmd_textbox.AppendText(txt)
+self.cmd_textbox.AppendText(txt.decode('GBk'))
than it work.
so,how can I solve the problem general,
modify my program or improve Gooey Library?

xavierskip:
I edit my program add this to solve this problem without modify the library file.
```
utf8_stdout = os.fdopen(sys.stdout.fileno(), mode='w', encoding='utf-8', closefd=False)
sys.stdout = utf8_stdout
```
via: https://stackoverflow.com/a/3597849/1265727

xavierskip:
if you redirect stdout when you package it with PyInstaller without console mode will raise issue, so i have to modify the library source code to pack it under windows system.

PR #237, opened 2017-12-31
set stdout encoding utf-8

xavierskip:
set this env to let it correct display result in Non-English language system。
my situation is in this issue: https://github.com/chriskiehl/Gooey/issues/230


chriskiehl:
This was fixed here: https://github.com/chriskiehl/Gooey/commit/4c61c29b650b2d43ca2476bf62eefdeb5dceb4ef#diff-e0df2c7c2df84db61e5994a80d1bc9dbR44

PR #241, opened 2018-01-29
Add "run_on_cli_callback" option to global configuration

crowjdh:
Some terminal-specific api calls(ex, curse.cbreak()) yields error and can't be adopted to current Gooey(wxPython) environment.
Receiving arguments from GUI and returning them to original process(mostly CLI) may resolve this issue, by supplying GUI just for getting arguments from user and calling callback function when "Start" button is clicked.

### DIFF
--- a/README.md
+++ b/README.md
@@ -246,7 +246,7 @@ Just about everything in Gooey's overall look and feel can be customized by pass
 | target | Tells Gooey how to re-invoke itself. By default Gooey will find python, but this allows you to specify the program (and arguments if supplied).|
 | program_name | The name displayed in the title bar of the GUI window. If not supplied, the title defaults to the script name pulled from `sys.argv[0]`. |
 | program_description | Sets the text displayed in the top panel of the `Settings` screen. Defaults to the description pulled from `ArgumentParser`. |
-| run_on_cli | Parse arguments from GUI and calls supplied function on cli(terminal) |
+| run_on_cli_callback | Parse arguments from GUI and calls this callback function on cli(terminal). The callback function takes two parameters(positional_args, optional_args) |
 | default_size | Initial size of the window | 
 | required_cols | Controls how many columns are in the Required Arguments section <br> :warning: **Deprecation notice:** See [Group Parameters](#group-configuration) for modern layout controls|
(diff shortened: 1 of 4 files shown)


crowjdh:
BTW, I'm glad you're back :)

chriskiehl:
Just to make sure I'm understanding correctly, could you give a super quick example of what would get passed to `run_on_cli_callback`

Thanks!

crowjdh:
I'm not sure why I forgot to add an example:

```python
import argparse

import time

def run_on_cli_callback(positional_args, optional_args):
  print "positional_args: %s, optional_args: %s" % (positional_args, optional_args)

@Gooey(run_on_cli_callback= run_on_cli_callback)
def main():
  parser = GooeyParser(description='Process some integers.')

  parser.add_argument(
    'required_field',
    metavar='Some Field',
    help='Enter some text!')

  parser.add_argument(
    '-f', '--foo',
    metavar='Some Flag',
    action='store_true',
    help='I turn things on and off')

  args = parser.parse_args()

if __name__ == '__main__':
  main()

```

Additionally, I updated README.md file to correct misspelled name of this option.

chriskiehl:
Ah, I see now. Thanks for the example. So, I think I misunderstood your original description. I thought the idea was to just dump the cli string generated by Gooey to stdout and then shut itself down. Basically just using Gooey as a way of getting input and then returning control back to the cli.

My current goal for Gooey making it completely language agnostic. As such, the whole thing is currently from a json file (which you can see via the `dump_build_config` option (also, a feature which your pull currently breaks ^_^)). The Gooey decorator and all the argparse stuff is just a convenience layer for generating that json config and spawning an instance of Gooey. Which is all to say, I don't want to introduce things into Gooey that aren't serializable and that can't exist outside of the Python world. 

Would the functionality you're after work _without_ the need for a custom CLI handling function? Or, if you're after the capability of having some kind of post-processing on the inputs, would having Gooey call your program with a specific flag work (a strategy I used for the [Dynamic Values](https://github.com/chriskiehl/Gooey#using-dynamic-values) functionality)?

crowjdh:
Oh, I see. What I've tried to achieve is to use cli-specific commands, which I wasn't able to use with Gooey.
I understand this feature is against your goal. I'd better use this feature from my forked repository ;)

PR #248, opened 2018-02-05
Add Hebrew language

ronyhe:
I added a json containing Hebrew messages. However, there seem to be two issues that require consideration:
A. Not all the messages are actually taken from the Hebrew json, some are still in English (screenshot attached
gooey
).
B. Hebrew is an RTL language, so full featured localization is a larger design issue.

P.S.
This is my first pull request EVER!
I'd love to get  feedback. Could something have been done better?


chriskiehl:
>This is my first pull request EVER!

Congrats! ^_^ 

This looks good to me! 

For the missing language bits in the dialog, would you mind opening an issue describing it as you did here? Or, if you want some bonus points, wanna fix it in a new pull request? It looks like I forgot to pass in the translated name to most of the [Chooser Types](https://github.com/chriskiehl/Gooey/blob/master/gooey/gui/components/widgets/core/chooser.py#L65). The docs say the Dialogs take that "Save a File" string in your picture as a first argument, so, it should be pretty straight forward to fix (you can see an example of the [translation function here](https://github.com/chriskiehl/Gooey/blob/master/gooey/gui/components/widgets/core/chooser.py#L44)) Make sure you update all of the language files, though! There's a `language_parity` test that'll help you out and make sure you didn't miss anything. 

Lemme know if you wanna take a stab at it. Otherwise, feel free to open an issue and I'll get around to it in a little while. 

Thanks for the pull!

ronyhe:
Sorry for the delay. 
I will open the issue and try to fix it in the coming days. 
Thanks!

PR #250, opened 2018-02-06
relax requirements.txt to allow for later versions

pmagwene:
Simple change to prevent pypi from stomping on required packages with later versions.


chriskiehl:
👍

Issue #249, opened 2018-02-05
Return back to Settings automatically after run without manually clicking Edit


dagrha:
Gooey is a great program-- thank you for open-sourcing it and for your work on it.

Is there a way to have the program return back to the "Settings" view after a successful run? In other words, after options are set in the Settings window and Start is clicked, the program runs and ends up at a "Finished" screen with an "Edit" button. If I click the Edit button it takes me back to the Settings view. Is there a way to automatically return to the Settings window without having to manually click Edit?

chriskiehl:
I can add one : )

dagrha:
Fantastic, thank for being willing to do that, Chris!

Is the solution a matter of "simulating" a click event? In other words, use wx.PostEvent() to "click" the Edit button programatically? Or something else?

chriskiehl:
If you're interested in doing it yourself (and making a pull request! 😄 ), the basic flow of adding a new option to Gooey is: 

1. Add it as an option in [config_generator.py](https://github.com/chriskiehl/Gooey/blob/master/gooey/python_bindings/config_generator.py).  
2. In [application.py](https://github.com/chriskiehl/Gooey/blob/master/gooey/gui/containers/application.py#L117) all that config data from config_generator is present on the instance as `self.buildSpec`. In this class there is a handler called `onComplete`. You'd just add a call to `showConfiguration()` within the `onSuccess` path if your flag is present (e.g. `if self.buildSpec.get('YOUR_OPTION', False): ...`).  Note: I'd keep the error path the same so that it doesn't silently fail in the case of an error 
3. Update the Configuration section of the README outlining your new option 
4. Profit  

Let me know if you want to take a stab at it.

dagrha:
I will take a stab at it. Thanks for the rundown on how to accomplish it. 👍

chriskiehl:
Closing. Fixed by your pull here: d01afc9b7be5637e7d4eb80ba49f11dbe0ebd8c7 

Nice work!

PR #251, opened 2018-02-07
add return_to_config option

dagrha:
This PR adds the feature discussed in #249

### DIFF
--- a/gooey/gui/containers/application.py
+++ b/gooey/gui/containers/application.py
@@ -123,7 +123,6 @@ class GooeyApplication(wx.Frame):
             if self.clientRunner.was_success():
                 if self.buildSpec.get('return_to_config', False):
                     self.showSettings()
-                    self.footer.progress_bar.Show(False)
                 else:
                     self.showSuccess()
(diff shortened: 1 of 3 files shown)


gooey/gui/containers/application.py
```diff
@@ -121,9 +121,13 @@ def onComplete(self, *args, **kwargs):
         """
         with transactUI(self):
             if self.clientRunner.was_success():
-                self.showSuccess()
-                if self.buildSpec.get('show_success_modal', True):
-                    wx.CallAfter(modals.showSuccess)
+                if self.buildSpec.get('return_to_config', False):
+                    self.showSettings()
+                    self.footer.progress_bar.Show(False)
```
chriskiehl: I'd put this line down in the showSettings() method. Other than that, lgtm.

dagrha: Do you mean, move the `self.footer.progress_bar.Show(False)` statement to the `showSettings()` method?

chriskiehl:
Awesome work! Thanks for the commit!

PR #262, opened 2018-02-20
fallback when decoding errors happen

jwaschkau:
I am using x86 Python 3.4 on Windows. The encoding for console output on englisch systems is cp1252 and not UTF-8. To get output when unexpected characters get decoded i added the 'replace' option sao the end user can see something. Maybe we need an option to set encoding for the ui and the console independently?

### DIFF
--- a/gooey/gui/processor.py
+++ b/gooey/gui/processor.py
@@ -63,7 +63,8 @@ class ProcessController(object):
             line = process.stdout.readline()
             if not line:
                 break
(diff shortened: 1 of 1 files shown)


chriskiehl:
>set encoding for the ui and the console independently? 

Yeah! Configure all the things! Wanna upgrade this pull request and implement that extra customization? ^_^ 

I think it would just be..

* a change in `config_generator.py` to add a new `ui_text_encoding` option
* passing that option to `ProcessController` rather than the default `encoding` that's there now

Plus... whatever I'm forgetting that would also be involved.

PR #271, opened 2018-03-01
Update chinese.json

machnine:
I have updated all 'translate me' and corrected a couple of odd expressions.


chriskiehl:
👍 Thanks!

PR #278, opened 2018-03-19
Compatibility with legacy versions of Pillow

saddy001:
Old versions of Pillow:Image don't support .tobytes(), so we use .tostring() instead.


chriskiehl:
lgtm. 👍

PR #279, opened 2018-03-23
Fix config load

sodwyer:
I was having issues when running my python scripts with specific python versions and with full file paths.
For example, something like: 
C:/Users/sodwyer/AppData/Local/Programs/Python/Python36-32\python.exe D:/Path/To/Project/proj_script.py

The JSON config file would fail to load, and the dumped config file would be dumped into the directory where python.exe was run from.

I made a couple of changes so that the folder of the current script is used instead. It seems to work ok for my use case but maybe it's not desired? Also, I didnt see any open issues about this so maybe it's a real edge case. Thought I'd pull request anyway just in case.

Let me know, thanks

### DIFF
--- a/gooey/python_bindings/gooey_decorator.py
+++ b/gooey/python_bindings/gooey_decorator.py
@@ -73,7 +73,7 @@ def Gooey(f=None,
           **params)
 
       if dump_build_config:
-        config_path = os.path.join(os.getcwd(), 'gooey_config.json')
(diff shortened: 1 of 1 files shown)


chriskiehl:
Long delay (sorry!), but these changes look good to me. Please re-open this against the [current release branch](https://github.com/chriskiehl/Gooey/tree/1.0.3-release)

sodwyer:
done

PR #283, opened 2018-04-11
Made Czech translation

cube181282s:
I was bored, so I made this translation for Czech. I would be happy if I helped this project at least little bit... 🙂
Note: I'm an native Czech speaker, so this translation should be very accurate.


chriskiehl:
Super cool. Thanks!

PR #291, opened 2018-05-10
Add MultiFileChooser to the README widgets table

jni:
It took me a while to figure out that this existed, and having it in this table would have helped!


chriskiehl:
Awesome. Thanks! 👍

Issue #293, opened 2018-05-28
Feature request: Allow general callbacks for validation


fuglede:
Currently, a `GooeyParser` can take a general string for validation, and this string will then be `eval`'ed when necessary. While this works for simple stuff, anything that requires more than Python built-ins becomes hard to express. For example, it could be natural to test `os.path.isfile` on the contents of a `FileChooser`, but that's not currently possible (since you would have to `import os` at some point).

Instead, it would be nice to be able to provide a completely general Python callable for use in validation. One possible API would be to simply allow the existing `TextContainer._options['validator']['test']` to be a callable and check its type when necessary. Otherwise, one monkey patch workaround is given in the following example in which we extend `TextContainer._options['validator']` to allow callables as well (but in the process break the ordinary validation):

```python
from gooey.gui.components.widgets.bases import TextContainer
oldGetValue = TextContainer.getValue

def newGetValue(self):
    result = oldGetValue(self)
    userValidator = self._options['validator']['callback']
    message = self._options['validator']['message']
    value = self.getWidgetValue()
    validates = userValidator(value)
    result['test'] = validates
    result['error'] = None if validates else message
    return result

TextContainer.getValue = newGetValue

@Gooey
def main():
    parser = GooeyParser()
    parser.add_argument('stuff', widget='FileChooser', gooey_options={
        'validator':{
            'callback': os.path.isfile,
            'message': 'some helpful message'
        }
    })
    args = parser.parse_args()
```

bilderbuchi:
You could probably have both ways by supplying the eval approach with a lambda as a default for a keyword argument that can be overridden with a user callable?

fuglede:
@bilderbuchi: So if no callable is provided in what I call `'callback'` in the example, it falls back to the contents of `'test'`, otherwise ignores it? If so, makes sense. Otherwise, what I was thinking was that if you wanted to keep it as a single option `'test'` (since you'd never use them at the same time anyway), you could, in [`TextContainer.getValue`](https://github.com/chriskiehl/Gooey/blob/b19fe8826dfae85c013506b4d49bbaee478e2006/gooey/gui/components/widgets/bases.py#L99), just do something to the extent of
```python
testFunc = userValidator if callable(userValidator) else eval(f'lambda user_input: bool({userValidator})')`
```

chriskiehl:
[...]

PR #307, opened 2018-07-24
293-TextContainer has new field callback.

Magnati:
Changed method in `gooey/gui/components/widgets/bases.py`:
I changed only `TextContainer.getValue `in the following way. 
	The fields `['validator']['callback']` is used to validate `validatesCB` with the `value` if existent and is set to  `True` otherwise.
	`result[error]` is set by an AND concatenation of both fields, `validatesCB` and the old `runValidator(satisfies, value) `.
	Like that `Gooey.TextContainer` becomes general callback applicable. And its behaviour does not change when no callback is present.

``` python  
def getValue(self):
        userValidatorCB = None
        validatesCB = True
        value = self.getWidgetValue()
        if 'callback' in self._options['validator']:
            userValidatorCB = self._options['validator']['callback']
            validatesCB = userValidatorCB(value)

        userValidator = getin(self._options, ['validator', 'test'], 'True')
        message = getin(self._options, ['validator', 'message'], '')
        testFunc = eval('lambda user_input: bool(%s)' % userValidator)
        satisfies = testFunc if self._meta['required'] else ifPresent(testFunc)
        return {
            'id': self._id,
            'cmd': self.formatOutput(self._meta, value),
            'rawValue': value,
            'test': runValidator(satisfies, value),
            'error': None if runValidator(satisfies, value) and validatesCB else message,
            'clitype': 'positional'
                        if self._meta['required'] and not self._meta['commands']
                        else 'optional'
        }
```


chriskiehl:
Closing this one as it's not inline with Gooey's direction. Detailed comments here: https://github.com/chriskiehl/Gooey/issues/293

PR #320, opened 2018-09-17
Update polish.json

rhkubiak:
I missed one English word.


chriskiehl:
Ah, good catch! Thanks again!

Issue #323, opened 2018-09-24
wxpython >=4.0.2 NewId depricated


eskemojoe007:
Hello There, Future Issue Creator! 

If you've found a bug, make sure to include the following:

 - [Windows 10 64 Bit] OS
 - [ 3.6.5] Python Version 
 - [ 1.0.1] Gooey Version 
 - `requirements.txt` states that any `wxpython` beyond `4.0.0b1` is acceptable, but wxpython `4.0.2` depricated `wx.NewID()` which is used in many places.  I get many deprectation warnings when running in `4.0.3` or `4.0.2`:

```
C:\Users\user\.virtualenvs\dialer_database-fp33vZHS\lib\site-packages\gooey\gui\events.py:14: DeprecationWarning: NewId() is deprecated
  WINDOW_RESTART  = wx.NewId()

C:\Users\user\.virtualenvs\dialer_database-fp33vZHS\lib\site-packages\gooey\gui\events.py:15: DeprecationWarning: NewId() is deprecated
  WINDOW_EDIT     = wx.NewId()

C:\Users\user\.virtualenvs\dialer_database-fp33vZHS\lib\site-packages\gooey\gui\events.py:17: DeprecationWarning: NewId() is deprecated
  WINDOW_CHANGE   = wx.NewId()
```

Either requirements.txt needs to be locked down or all references to `wx.NewId()` need to be replaced with `wx.NewIdRef()`.  https://groups.google.com/forum/#!topic/wxpython-users/r7Q5pFge2KE

chriskiehl:
Resolved by https://github.com/chriskiehl/Gooey/pull/324

PR #324, opened 2018-09-24
Update to wxpython >=4.0.2 #323

eskemojoe007:
Fixing #323 in order to accommodate the new `wxpyton` versions and deprications of `wx.NewId()`

### DIFF
--- a/gooey/gui/events.py
+++ b/gooey/gui/events.py
@@ -7,20 +7,19 @@ that tie everything together.
 
 import wx
 
-WINDOW_STOP     = wx.NewId()
-WINDOW_CANCEL   = wx.NewId()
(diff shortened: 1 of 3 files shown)


nelsonjchen:
Are there any blockers to this? It's a bit annoying to see my unit tests keep warning about the usage of `NewID`. I can lock down my application's `requirements.txt` to 4.0.1 and downgrade to silence the warnings but I don't think that's ideal either.

nelsonjchen:
Alternatively, would a PR for a version where `NewID` is used on `4.0.1` but `NewIDRef` is used on > 4.02 be more acceptable?

cclauss:
It seems like these changes were reverted and now pytest raises a lot of warnings.

PR #331, opened 2018-10-01
Avoid importing wx when --ignore-gooey is part of the command line

NathanRichard:
# The problem this PR is trying to solve
I built an application that uses Gooey to generate its user interface. It has been working very well on my development machine, both in gui and cli mode. However, when I tried to use the application on a headless server, I got the following error:
```
ImportError: libgtk-3.so.0: cannot open shared object file: No such file or directory
```
Of couse, gtk is not installed on this server and I did not expect to use any of the graphical features of gooey on this machine. But as gooey allows the applications to be used in cli mode (using the --ignore-gooey switch), I would expect it to gracefully degrade its functionalities and require the graphical backend only when the graphical mode is used. This would allow client applications to be more versatile and better adapt to their environment.

# The PR content

The PR itself is quite simple, I removed the transitive inclusion of wx from the --ignore-gooey codepath and lazily imported it at the beginning of the pure graphical codepath.

Even before my change I did not manage to run the tests, but all my application tests depending on gooey work well and it solved the gtk issue.

I am not used to doing PR to python projects, so if this PR breaks some guideline or good pratice, please tell me, I will be glad to correct it.

### DIFF
--- a/gooey/python_bindings/gooey_decorator.py
+++ b/gooey/python_bindings/gooey_decorator.py
@@ -11,7 +11,6 @@ import os
 import sys
 from argparse import ArgumentParser
 
-from gooey.gui import application
 from gooey.gui.util.freeze import getResourcePath
 from gooey.util.functional import merge
 from . import config_generator
@@ -53,6 +52,7 @@ def Gooey(f=None,
 
   def build(payload):
     def run_gooey(self, args=None, namespace=None):
+      from gooey.gui import application
       source_path = sys.argv[0]
 
       build_spec = None


gooey/python_bindings/gooey_decorator.py
```diff
@@ -53,6 +52,7 @@ def Gooey(f=None,
 
   def build(payload):
     def run_gooey(self, args=None, namespace=None):
+      from gooey.gui import application
```
chriskiehl: Could you add a quick one-liner comment explaining why the import is deferred (so I don't blindly move it back out at some later date)

chriskiehl:
This change lgtm! Please re-open it against the current [release branch](https://github.com/chriskiehl/Gooey/tree/1.0.3-release)?

PR #343, opened 2018-10-30
Change "choose one" to "choose_one" in languages

eturkes:
The key is incorrectly specified which causes the mutually exclusive
group radio buttons to display "choose_one" instead of the proper
language value like "Choose One".


chriskiehl:
👍 thanks!

PR #352, opened 2018-11-15
fix listbox together with poll_external_updates=True

gonzygonz:
Fix ListBox together with poll_external_updates=True


chriskiehl:
Thanks for fixing this! The next release in the works is https://github.com/chriskiehl/Gooey/tree/1.0.2-release. Would you mind updating this PR to target that branch?

gonzygonz:
ok
done with #353 
will close this one

PR #362, opened 2018-12-21
Add parameter clear_before_run that forces console clear between runs.

NathanRichard:
If the parameter is set to yes in the gooey decorator, the console will be cleared before each run.

Current default parameter value is conservative and does not clear the console.


chriskiehl:
👍

PR #365, opened 2019-01-03
Handling of style-oriented ascii espace sequences.

NathanRichard:
# Motivation
I wanted my gooey-powered programs to be able to display rich text (bold, colors, etc.) the same way that they do when running in a normal terminal emulator.

# Description of the change

Using the RichTextCtrl widget we are able to use advanced formatting in the console panel. The formatting information is extracted from the text stream by searching for terminal escape sequences. The sequence values are taken from the colored package, which becomes a lazy dependency (imported only when the feature is activated).

The previous implementation (using the TextCtrl widget) has been extracted into a separate class.

The client code chooses between the basic or the rich class based on the richtext_controls configuration parameter. The default is to use the previous implementation based on TextCtrl.

# Note to reviewer

This change is much bigger that any other that I submitted to the Gooey repo. It may conflict with your vision of the project or some project guideline I wasn't aware of. If it does, I will be happy to have your feedback and improve this PR.

### DIFF
--- a/gooey/python_bindings/config_generator.py
+++ b/gooey/python_bindings/config_generator.py
@@ -86,6 +86,7 @@ def create_from_parser(parser, source_path, **kwargs):
       'terminal_font_family': kwargs.get('terminal_font_family', None),
       'terminal_font_weight': kwargs.get('terminal_font_weight', None),
       'terminal_font_size':   kwargs.get('terminal_font_size', None),
+      'richtext_controls':    kwargs.get('richtext_controls', False),
       'error_color':          kwargs.get('error_color', '#ea7878')
   }
 
--- a/README.md
+++ b/README.md
@@ -303,6 +303,7 @@ Just about everything in Gooey's overall look and feel can be customized by pass
 | terminal_font_weight | Weight of the font (NORMAL|BOLD) | 
 | terminal_font_size | Point size of the font displayed in the terminal | 
 | error_color | HEX value of the text displayed when a validation error occurs |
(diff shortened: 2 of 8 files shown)


chriskiehl:
**Note to implementer:**

* Project vision conflicts: none
* project guidelines violated: none

This has a thorough description of what its doing, maintains backwards compatibility, and defaults to the old stuff. This is a solid PR, friendo. 

Do you have an example for schlubs like me that dunno how to actually color things on the terminal? Such a thing would also make for a stellar contribution to the [Gooey Examples](https://github.com/chriskiehl/GooeyExamples) repo, btw (hint hint)!

gooey/python_bindings/config_generator.py
```diff
@@ -86,6 +86,7 @@ def create_from_parser(parser, source_path, **kwargs):
       'terminal_font_family': kwargs.get('terminal_font_family', None),
       'terminal_font_weight': kwargs.get('terminal_font_weight', None),
       'terminal_font_size':   kwargs.get('terminal_font_size', None),
+      'richtext_controls':    kwargs.get('richtext_controls', False),
```
chriskiehl: Oh, make sure to add this to the README

NathanRichard: As requested, I added an entry in the README.md describing the new feature.
I also created a simple example in the examples repo and opened a PR for it :
https://github.com/chriskiehl/GooeyExamples/pull/16

chriskiehl:
👍

PR #367, opened 2019-01-09
Allow hiding progress regex stdout from console

conradhilley:
I have added a hide_progress_msg argument to the gooey decorator. If this 'hide_progress_msg' parameter is set to True , stdout which satisfies the progress_regex will only be used to update the progress bar and not be printed to the console. The current default parameter retains existing behavior. Let me know if there is a different name you'd like for the argument.

I also updated a couple imports in the tests and menubar.py to be absolute, however I'm not sure if those will cause any issues with Python 2.

### DIFF
--- a/gooey/gui/components/menubar.py
+++ b/gooey/gui/components/menubar.py
@@ -3,7 +3,7 @@ from functools import partial
 
 import wx
 
-from gui import three_to_four
+from gooey.gui import three_to_four
 
 
 class MenuBar(wx.MenuBar):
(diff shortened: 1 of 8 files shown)


chriskiehl:
Cool. I dig it! 

Changes look fine to me. Would you mind updating the README with your new option?

conradhilley:
Added to the Global Configuration and Showing Progress sections of the README. Later I can add an example to the Gooey Examples repository.

Issue #377, opened 2019-02-05
[TypeError: cannot serialize '_io.TextIOWrapper' object] when using [metavar='<FILE>', type=argparse.FileType()].


changeling:
- OSX Mojave
 - Python 3.7.2
 - Gooey 1.0.2
 - Expected behavior: Handling of FileType. 
 - Actual Behavior: `TypeError: cannot serialize '_io.TextIOWrapper' object`

 - Minimal code example:
```
#!/usr/bin/env python3
# -*- coding: utf-8 -*-
"""
Minimal example for [metavar='<FILE>', type=argparse.FileType()] issue.

NOTE: Same behavior when `GooeyParser` replaces `argparse.ArgumentParser`.
NOTE: Behavior is independent of `FileType()` arguments:
        `FileType('w', encoding='UTF-8')` and
        `FileType()` yield same failure.
"""

# global import
import sys
import argparse

try:
    from gooey import Gooey
except ImportError:
    sys.stderr.write('You need to install the gooey module first\n')
    sys.stderr.write('("python3 -m pip install gooey"\n')
    exit(2)


@Gooey
def main():
    """Demonstrate [metavar='<FILE>', type=argparse.FileType()] issue."""
    parser = argparse.ArgumentParser(
        description="Show [metavar='<FILE>', type=argparse.FileType()] issue.",
        add_help=False,
        usage="test.py -u username -p password [options]")
    parser.add_argument('-u',
                        metavar='<STR>', type=str, help='Username')
    parser.add_argument('-p', metavar='<STR>', type=str, help='Password')
    try:
        parser.add_argument('-o',
                            metavar='<FILE>',
                            type=argparse.FileType('w', encoding='UTF-8'),
                            default=sys.stdout,
                            help='output data file [stdout]')
        parser.add_argument('-l',
                            metavar='<FILE>',
                            type=argparse.FileType('w', encoding='UTF-8'),
                            default=sys.stderr,
                            help='output log file [stderr]')
    except TypeError:
        sys.stderr.write(
            'Python >= 3.4 is required to run this script\n')
        sys.stderr.write(
            '(see https://docs.python.org/3/whatsnew/3.4.html#argparse)\n')
        exit(2)

    args = parser.parse_args()


if __name__ == '__main__':
    main()
```

changeling:
Digging in to try and get my head around this, I learned that deepcopy has issues with complex objects. The following seems directly on target:

[...]

PR #386, opened 2019-03-10
Update argparse_to_json.py for FileType actions.

changeling:
Add elif is_file(action) to set widget to FileSaver for FileType actions.
Add is_file() function.
Exclude FileType actions from is_standard().

### DIFF
--- a/gooey/python_bindings/argparse_to_json.py
+++ b/gooey/python_bindings/argparse_to_json.py
@@ -255,6 +255,9 @@ def categorize(actions, widget_dict, options):
 
         elif is_standard(action):
             yield action_to_json(action, _get_widget(action, 'TextField'), options)
+        
+        elif is_file(action):
+            yield action_to_json(action, _get_widget(action, 'FileSaver'), options)
 
         elif is_choice(action):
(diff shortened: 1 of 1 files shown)


changeling:
Welp. You know the old saying. Haste makes a mess. Apologies for the previous PR spaghetti. This should do the trick.

Re-adding note regarding https://github.com/chriskiehl/Gooey/issues/377 and its blocking of FileType actions for reference.

chriskiehl:
No worries -- happens all the time ^_^ 

This lgtm. Just one thing, would you mind deleting this line in the same file as part of your pull? https://github.com/chriskiehl/Gooey/pull/386/files#diff-c2a9e315e78b26f3c7b03697a9f06437R279 

It was trying (and failing) to do what your PR is doing. I apparently didn't know how instance checking worked x_x

changeling:
OK, how's that look now?

(And tbh, I would have expected that `==` to have worked as well. Learning something every day. Whee!)

chriskiehl:
Looks good. Thanks for the fix! 👍

PR #387, opened 2019-03-10
Remove leftover reference to `type_arg_widget`.

changeling:
argparse_to_json.py:

    return supplied_widget or type_arg_widget or default


changeling:
Um...

Whoops. :}

Issue #363, opened 2018-12-30
Wildcards in choosers


gonzygonz:
Hi there,
I saw there is a TODO to add wildcards to choosers as they are not supported at the moment.
i assume this issue is known so i didnt add any example. 

I think i know how to implement this feature, so if no one is working at it at the moment ill go ahead and prepare a pull request (and also support message and defaultFile along the way)

Thanks! :>

gonzygonz:
Hi,
I implemented wildcards support, and on the way also support for defaultFile, defaultFolder and message For choosers that support this.

I want to prepare a PR for this.
Based on which branch should i do it?

Thanks :>

chriskiehl:
Can't wait to check it out! 1.0.3-release is the chap you should target.

gonzygonz:
hi, is there anything still needed to accept the PR?

chriskiehl:
Nope. Just needed to poke me ^_^ Nights and weekends are eaten up by other projects these days.. 

I'll try to get all the pending PRs merged/reviewed soon. I'll try to get to them all over the weekend. And, then, hopefully, _finally_, actually kick 1.0.3 out the door!

gonzygonz:
Yea, i know the feeling.
Great job on this lib. Very useful for many small things i do. Keep up the good job
Cant wait to port to 1.0.3

chriskiehl:
Closing. Completed by you! ^_^

PR #400, opened 2019-03-31
Add support for wildcards, message, defualtFile, DefaultDir and DefatltPath for Choosers

gonzygonz:
#363
adding this feature
i had to delete previous PR as i did some mess there (again, im new to git and github)
i did the changes you suggested to use .get() method
now its a much cleaner implementation.

let me know if you need anything else

P.S after this PR is done, i will find time to update the example to use this new feature


gooey/gui/components/widgets/core/chooser.py
```diff
@@ -64,40 +64,57 @@ def getValue(self):
...
 class FileSaver(Chooser):
     """ Specify the path to save a new file """
     def getDialog(self):
+        options = self.Parent._options
         return wx.FileDialog(
             self,
             style=wx.FD_SAVE | wx.FD_OVERWRITE_PROMPT,
-            defaultFile=_("enter_filename"),
-            message=_('choose_file')
+            defaultFile=options.get('defaultFile', _("enter_filename")),
```
chriskiehl: These option keys should actually be snake_cased as they come from Gooey's public options API, which honors PEP8.

gooey/gui/components/widgets/core/chooser.py
```diff
@@ -64,40 +64,57 @@ def getValue(self):
 class FileChooser(Chooser):
     """ Retrieve an existing file from the system """
     def getDialog(self):
-        return wx.FileDialog(self, _('open_file'), style=wx.FD_OPEN | wx.FD_FILE_MUST_EXIST)
+        options = self.Parent._options
+        return wx.FileDialog(self, message=options.get('message', _('open_file')),
+                             style=wx.FD_OPEN | wx.FD_FILE_MUST_EXIST,
+                             defaultFile=options.get('defaultFile', _("enter_filename")),
```
chriskiehl: You'll need to add these new options to the Gooey Options docs here: https://github.com/chriskiehl/Gooey/blob/1.0.3-release/docs/Gooey-Options.md#individual-widget-options

gooey/gui/components/widgets/core/chooser.py
gonzygonz: hi
i kept options as they appear in original wx lib so users will have the same interface in Gooey as they have in wx.
besides - its a string dict key, so i dont think pep8 has any case enforce on this
(make sure you are not confused by the removed code which had variables and they really should have been snake_case)

i think its smarter to have it the same as wx, but if you feel like it should be snake_case i will change that.

chriskiehl: As much as possible, we shouldn't be exposing wx internals to outside clients or using it as a rational for driving APIs. Yes, we happen to be using the same keyname as wx here. Their names are Good Enough that it doesn't matter -- `default_dir` has no coupling to wx specifically -- we could rewrite Gooey with Tkinter or PyQt and use the reuse the name. Additionally, your options would be the _only_ options in Gooey's public API that are camelCased. That alone should be reason enough to snake_case.

gonzygonz:
Hi
i updated to snake_case and added documentation.
first time i do github md documentation so ill be happy for notes (if any)

eladeyal-intel:
@gonzygonz I propose that the docuentation can be clearer about when to use "default_path" and when to use "default_dir" and "default_file".

Also I propose not to use `message` as a parameter name. I think it the term is too broad to be useful. Perhaps `prompt` or `dialog_prompt` is better?

eladeyal-intel:
@gonzygonz `title` should be used in my opinion only for the text which is in the window title bar. If the text in question is beneath the title bar (of if the window is modal and there is no title bar) then I propose `prompt`

gooey/gui/components/widgets/core/chooser.py
gonzygonz: I've added documentation.
Please review it is clear.

chriskiehl:
Long delay, but looks good! Thanks for the addition!

Issue #408, opened 2019-04-11
hide option from gooey


eladeyal-intel:
I want to "hide" an option from Gooey; so it's visible when running the python script with -h but not controllable via GUI.

It can be useful for options which are too sophisticated to be controlled via GUI.

Maybe easy to implement by adding a "Hidden" widget

gonzygonz:
Hi 
I didnt try it yet, but i think you can easily do it woth regular argparse before the gooey argparse.
Just make another argparse before, which knows the new extra options, and argparse can return the arga from there, and the rest of params, and just pass the "reat" options into the gooey argparse. This way you filter out the unwanted options. 

Again, i didbt test ot yet, but if this explanation is not clear i can cook some example later on next week.

Hope this helps

gonzygonz:
hi again, 
so i tried it but its not really trivial for a common user to be able to do that.
so maybe this hidden option is really something to consider.
i am not sure if it should be a widget or a gooey_option

chriskiehl:
Implemented on the [release branch](https://github.com/chriskiehl/Gooey/commit/d97924bfb4323db4b62ac34c2b2e6a1cc5ac2019). Give it a whirl and lemme know if you find any quirks! 

Note: this is a true _visibility_ toggle. The widget is still there, and will ship whatever default you provide it (when a required argument). 

It's controlled via `gooey_options` rather than a custom widget, as they already were collecting a bunch of individual `x_visible` props. 

```
from gooey import Gooey, GooeyParser


@Gooey
def main():
    parser = GooeyParser()
    parser.add_argument('filename1')
    parser.add_argument(
        'secret-advanced-argument',
        default='TOO POWERFUL!!',
        gooey_options={'visible': False}
    )
    args = parser.parse_args()


if __name__ == '__main__':
    main()
```

PR #419, opened 2019-04-22
README

eladeyal-intel:
sentence was cut.

also: describe #408


chriskiehl:
README is getting crazy long. I got rid of this section and moved it into a `/docs` folder (which I'll slowly start populating...). 

I did totally forget about #408 that you mention here, so if you wanna patch the new docs file with that info it'd be super duper. Otherwise, I opened an [issue](https://github.com/chriskiehl/Gooey/issues/424) to remind myself to do it later

PR #430, opened 2019-05-19
detect URLs in rich console and make them clickable

### LINKED ISSUE #421: Auto-URL for rich console
@NathanRichard maybe you want to help me with this.

In https://github.com/chriskiehl/Gooey/commit/f581884fbf86ecb0634c91312cb9dab86c0d2281 and https://github.com/chriskiehl/Gooey/pull/417 we (well @chriskiehl  ) added the auto-URL, so the console identifies a URL as a clickable thing.

This doesn't work for rich controls. So the following code identifies the link if `richtext_controls=False` but not if it's `True`.

I tried to add `| wx.TE_AUTO_URL | wx.TE_RICH` to https://github.com/chriskiehl/Gooey/blob/1.0.3-release/gooey/gui/components/widgets/richtextconsole.py#L41 but it's not enough, things still don't work. 

```
import sys

@Gooey(richtext_controls=True,auto_start=True)
def main():
    _ = GooeyParser().parse_args(sys.argv[1:])
    print ('please visit http://google.com')

if __name__ == '__main__':
    main()
```

### LINKED ISSUE #427: Auto-URL is confused by wide characters
related to #404 

```
import sys

@Gooey(richtext_controls=False,auto_start=True)
def main():
    _ = GooeyParser().parse_args(sys.argv[1:])
    print ('please 📧📧📧📧 visit http://google.com because life is pretty')

if __name__ == '__main__':
    main()
```

It detects the URL but cliking it tries to go to `http://google.com%20bec` 

So the four extra-wide characters caused the URL to end (and I suspect also: start) 4 character too late. 

I imagine it's something with `event.URLStart:event.URLEnd` counting bytes instead of characters or vice versa,  in  https://github.com/chriskiehl/Gooey/blob/1202f13a5c916a5a4c07590a8c55b523de7cf606/gooey/gui/components/console.py#L43 

tested in python 3 only so far.

eladeyal-intel:
Closes #421 - we detect URLs using a regex and applying the URL style.

Also closes #427 becasue it's the same line

### DIFF
--- a/gooey/gui/components/widgets/richtextconsole.py
+++ b/gooey/gui/components/widgets/richtextconsole.py
@@ -1,6 +1,7 @@
 import wx
 import wx.richtext
 import colored
+import re
 
 kColorList = ["#000000", "#800000", "#008000", "#808000", "#000080", "#800080", "#008080", "#c0c0c0",
     "#808080", "#ff0000", "#00ff00", "#ffff00", "#0000ff", "#ff00ff", "#00ffff", "#ffffff", "#000000",
@@ -39,6 +40,8 @@ class RichTextConsole(wx.richtext.RichTextCtrl):
 
     def __init__(self, parent):
         super(wx.richtext.RichTextCtrl, self).__init__(parent, -1, "", style=wx.richtext.RE_MULTILINE | wx.richtext.RE_READONLY)
+        self.regex_urls=re.compile(r'\b((?:file://|https?://|mailto:)[^][\s<>|]*)')
+        self.url_colour = wx.Colour(0,0,255)
         self.esc = colored.style.ESC
         self.end = colored.style.END
         self.noop = lambda *args, **kwargs: None
@@ -58,6 +61,25 @@ class RichTextConsole(wx.richtext.RichTextCtrl):
             # NB : we use a default parameter to force the evaluation of the binding
             self.actionsMap[escSeq] = lambda bindedColor=wxcolor: self.BeginTextColour(bindedColor)
 
+    def PreprocessAndWriteText(self, content):
+        """Write text into console, while capturing URLs and making 
+        them blue, underlined, and clickable.
+        """
+        textStream=iter(re.split(self.regex_urls, content))
+        # The odd elements in textStream are plaintext;
+        # the even elements are URLs.
+        for plaintext in textStream:
+            url=next(textStream, None)
+            self.WriteText(plaintext)
+            if url:    
+                self.BeginTextColour(self.url_colour)
+                self.BeginUnderline()
+                self.BeginURL(url)
+                self.WriteText(url)
+                self.EndURL()
+                self.EndUnderline()
+                self.EndTextColour()
+            
     def AppendText(self, content):
         """
         wx method overriden to capture the terminal control character and translate them into wx styles.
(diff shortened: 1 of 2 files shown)


eladeyal-intel:
Wait... Don't merge

gooey/gui/components/widgets/richtextconsole.py
```diff
@@ -59,6 +61,27 @@ def __init__(self, parent):
             # NB : we use a default parameter to force the evaluation of the binding
             self.actionsMap[escSeq] = lambda bindedColor=wxcolor: self.BeginTextColour(bindedColor)
 
+    def AppendText2(self, content):
```
chriskiehl: Is there a better name that AppendText2?

gooey/gui/components/widgets/richtextconsole.py
```diff
@@ -59,6 +61,27 @@ def __init__(self, parent):
...
+        """Capture URLs and make them blue, underlined, and clickable."""
+        while True:
```
chriskiehl: I don't dig this while loop + mutation of the content variable. 

As a possible alternative, if you expand your regex's capture group to include the whole match, you can simplify this whole thing to a single forloop + if statement

```python 
url_pattern = r'\b((?:file|https?)://[^\s<>]*)'
```

>note that we treat the inner scheme subgroup as non-capturing 

then re.split will give you stable pairs of [plaintext, url]

```python
re.split(r, 'http://www.example.com')
>>> ['', 'http://www.example.com', '']
re.split(r, 'Hello world!\n\nhttp://www.example.com')
>>> ['Hello world!\n\n', 'http://www.example.com', '']
re.split(r, 'file://www.example.com')
>>> ['', 'file://www.example.com', '']
re.split(r, 'some text file://www.example.com some more text http://www.example.com http://www.example.com')
>>> ['some text ', 'file://www.example.com', ' some more text ', 'http://www.example.com', ' ', 'http://www.example.com', '']
```

Combine this with `itertools.grouper` and you can iterate over the pairs directly

```python
for plaintext, url in grouper(re.split(pattern, content), 2):
    self.WriteText(plaintext)
    if url:
        self.BeginTextColour(self.url_colour)
        self.BeginUnderline()
        self.BeginURL(url)
        self.WriteText(url)
        self.EndURL()
        self.EndUnderline()
        self.EndTextColour()
```

eladeyal-intel: `grouper` is not actually in itertools, it's just a "recipe" in https://docs.python.org/3.7/library/itertools.html#itertools-recipes ; and it needs either `izip_longest` or `zip_longest` (for Python 2 vs 3) which seems to me like a hassle. I hope the updated version is more to your liking.

gooey/gui/components/widgets/richtextconsole.py
```diff
@@ -58,6 +61,24 @@ def __init__(self, parent):
             # NB : we use a default parameter to force the evaluation of the binding
             self.actionsMap[escSeq] = lambda bindedColor=wxcolor: self.BeginTextColour(bindedColor)
 
+    def PrepocessAndWriteText(self, content):
+        """Write text into console, while capturing URLs and making 
+        them blue, underlined, and clickable.
+        """
+        l=re.split(self.regex_urls, content)
+        while l:
```
chriskiehl: I still feel like you're over thinking this a bit with the while loop + manual deletions of list items as you consume them. Iterators can handle all this for you (without the hefty O(n) tax for those list deletions) 

```
    textStream = iter(re.split(r, x))
    for item in textStream:
        plaintext = item
        url = next(textStream, None)
        # no additional bookkeeping required
        write(plaintext)
        if url:
            write(url)
```

eladeyal-intel: done. also added mailto: URLs while at it

chriskiehl:
👍

PR #431, opened 2019-05-19
Auto-URLs are confused by wide characters - close #427

### LINKED ISSUE #427: Auto-URL is confused by wide characters

eladeyal-intel:
wxpython is explicit that you can't mix GetValue() and positions you get from other APIs.

### DIFF
--- a/gooey/gui/components/console.py
+++ b/gooey/gui/components/console.py
@@ -40,7 +40,7 @@ class Console(wx.Panel):
 
     def evtUrl(self, event):
         if event.MouseEvent.LeftUp():
-            webbrowser.open(self.textbox.GetValue()[event.URLStart:event.URLEnd])
+            webbrowser.open(self.textbox.GetRange(event.URLStart,event.URLEnd))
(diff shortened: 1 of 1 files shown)


chriskiehl:
Closing in favor of the other CR.

Issue #463, opened 2019-08-09
Release 1.0.3, Help text is destroyed.


thorstenwagner:
- [x] OS Kubuntu 18.04
 - [x] Python Version 3.6
 - [x] Gooey Version 1.0.3-release

I tried 1.0.3-release as long help texts are cropped in 1.0.2. Unfortunately the help text are now completely destroyed :-)

Here an example to reproduce:
 ```

from gooey import Gooey, GooeyParser

def main():
    parser = GooeyParser(
        description="A GooeyParser example",
       # formatter_class=argparse.ArgumentDefaultsHelpFormatter,
    )
    required_group = parser.add_argument_group(
        "Required arguments",
        "These options are mandatory "
    )

    required_group.add_argument("-a", "--a_argument", required=True, help="This is a argument.")

    required_group.add_argument(
        "-b",
        "--b_argument",
        type=int,
        required=True,
        default=5,
        help="This is a long text. This is a long text. This is a long text. This is a long text. ",
    )
    parser.parse_args()

def _main_():

    Gooey(main,
          program_name="Reproduce something",
          tabbed_groups=True)()

if __name__ == "__main__":
    _main_()
```

The result looks like this:
![gooey_problem](https://user-images.githubusercontent.com/1113977/62769905-36491580-ba9a-11e9-9ae2-278267f316a1.png)

Additionally, the start and cancel buttons are always cutted. This is  the case in 1.0.2 as well.
Any ideas?

Thanks!
Thorsten

thorstenwagner:
I've found a fix, could you have a look?

chriskiehl:
Fix lgtm! Will merge this weekend

chriskiehl:
Closing. Fixed by you!

PR #464, opened 2019-08-09
Fix text wrapping in 1.0.3-release

thorstenwagner:
This fixes the problem reported in https://github.com/chriskiehl/Gooey/issues/463


chriskiehl:
Looks good on my end! Nice work!

Issue #297, opened 2018-06-13
UI glitch under Ubuntu 16.04


gsemet:
Hello !

I have some glitch on Ubuntu 16.04 (Unity X11):
![gooey-glitch](https://user-images.githubusercontent.com/133498/41348790-14b9959e-6f0e-11e8-9d80-69fc7c35ef00.png)

Do you want what I can do to fix the button ? And maybe a slightly bigger scrollbar on the right?

Thanks

chriskiehl:
Fixed via 62643ae

PR #471, opened 2019-09-16
Change default button size for footer

thorstenwagner:
Version: 1.0.3
OS: KUbuntu

I've changed default button size of the footer buttons as they are shown incomplete on my system.

Before change:
button_before

After change:
button_after

### DIFF
--- a/gooey/gui/components/footer.py
+++ b/gooey/gui/components/footer.py
@@ -122,7 +122,7 @@ class Footer(wx.Panel):
         return wx.Button(
             parent=self,
             id=event_id,
-            size=(90, 24),
+            size=(90, 30),
             label=label,
             style=style)
 


gooey/gui/components/footer.py
```diff
@@ -122,7 +122,7 @@ def button(self, label=None, style=None, event_id=-1):
         return wx.Button(
             parent=self,
             id=event_id,
-            size=(90, 24),
+            size=(90, 30),
```
chriskiehl: Hm.. I'm not sure why I hard coded that height to begin with.. but if you set it to `-1` it should pick an appropriate height for all platforms (e.g. `size=(90, -1)`).

chriskiehl:
Fixed via https://github.com/chriskiehl/Gooey/commit/62643aec68f7992efbab2d1fe27948c20c495d9f 

Sorry! Trying to get everything out the door, and this is a good catch and closes down #297 - a long standing bug, so I made the swaperoo myself.

thorstenwagner:
Perfect, thanks for the fix!

PR #487, opened 2019-09-28
Korean and Japanese translate

partrita:
I fix the translation file.


chriskiehl:
could you update this to point at the 1.0.4-release branch? Thanks!

Issue #472, opened 2019-09-23
Print output


mathiasrw:
Feature request: 

Let the user click to print the output from running the command. Combined with a PDF printer this would be a great way of letting the user document something has been done.

PR #502, opened 2019-11-17
Implemented printing on success (Issue #472)

tom-p-reichel:
Used Gooey for a project and wanted to contribute, so I added a print button to the success screen as per some user requests. :smiley:
Tested on Python 3.8, Arch Linux 5.3.8.

The translations for "print" I yanked from google translate, using the term "print out" from English to those languages. I can tell that it produced a verb in Spanish, French, and Portugese, so those translations are likely fine. The rest may be of questionable quality.


chriskiehl:
Hey @tom-p-reichel, this is really nifty!

However, the **bad news**: I don't love the functionality as a button.. but the **good news**(!): I think this is a great fit for the menu system and would love to see it implemented!  

If you're up for it, checkout [menubar.py](https://github.com/chriskiehl/Gooey/blob/master/gooey/gui/components/menubar.py). You should be able to migrate your logic to that area. And if you go that route, don't forget to add your option to the README.

PR #380, opened 2019-02-27
Implementation of TimeChooser widget.

NathanRichard:
This is the implementation of the TimeChooser widget on the basis of DateChooser.
DateChooser was refactored so that TimeChooser and DateChooser share most of their code.

image

PR #504, opened 2019-11-19
Implement a TimeChooser Widget

NathanRichard:
This is another try at [one of my previous PR](https://github.com/chriskiehl/Gooey/pull/380). At the time I was not available enough to implement the comments made on the PR.
I hope this one will be ok.

# Changes
- Refactoring on the Calendar dialog to be a base class for both the date chooser and time chooser dialogs.
- Implementation of TimeChooser, and mention of this new widget in the README.
- Implementation of review comments from the first PR, including proper name casing and Python2 compatibility.
- Second batch of improvements for localization of these dialogs and style homogeneity with FileChooser (notably moving the label into the windows title).

# Related PR
Added TimeChooser to the widget_demo:
https://github.com/chriskiehl/GooeyExamples/pull/22


gooey/gui/components/widgets/dialogs/base_dialog.py
```diff
@@ -0,0 +1,54 @@
+from gooey.gui.lang.i18n import _
+
+__author__ = 'Chris'
```
chriskiehl: You can cut that bit

gooey/gui/three_to_four.py
```diff
@@ -26,6 +26,11 @@ class Classes:
     else:
         DatePickerCtrl = wx.DatePickerCtrl
 
+    if isLatestVersion:
```
chriskiehl: Nice. Good catch on this

chriskiehl:
New component is cool! 

Could you squash your commits down to a single feature commit? Once that's done and the time issue above addressed, looks good to my eye.

NathanRichard:
> Can you same-ify the time representations? It's a little jarring to go from a 12 hour clock to a 24 one!
> 
> image


I investigated the issue  and found no easy solution :
I was only able to reproduce the formatting discrepancy in windows using En_US locale (which uses the "h:mm:ss tt" format) :
image

For instance the English and French locales use the "H:mm:ss" format which is consistent with the ISO format used in the widget:
image

It is actually WxPython which decides to adapt to the locale, and the [wx TimePicker widget](https://wxpython.org/Phoenix/docs/html/wx.adv.TimePickerCtrl.html) does not allow to control that.

On gooey code, we force the formatting to ISO:
```python
class TimeDlg(BaseDialog):
	def __init__(self, parent):
		super(TimeDlg, self).__init__(parent, 
			pickerClass=Classes.TimePickerCtrl,
			pickerGetter=lambda datepicker: datepicker.GetValue().FormatISOTime(),
			localizedPickerLabel=_('select_time'))
```


We could use the [wx.DateTime.FormatDate()](https://wxpython.org/Phoenix/docs/html/wx.DateTime.html#wx-datetime) method to have it in the current locale instead, but given the current BaseDialog architecture, that would mean sending the time in the current locale to the client application as well.

So as I see things we have the following options :
1.   Always use the current locale and let the client application parse the value. That is both easy and user-friendly, but from the client application perspective that adds a lot of complexity ; there could be corner cases where WxPython and the client application (for instance using babel) do not agree on what the locale format should be, which would break the app for some users in conditions difficult to reproduce for the dev.
2. Refactor the BaseDialog class to be able to present the data in the gui in one format (the locale) and send it to the application in another. I am not sure it is worth the added complexity.
3. Keep it as it is (force ISO format in the widget and the value sent to the app). It may be a little disturbing for some users, but it provides a simpler contract to the client app and is more consistent with the other widgets (like CalendarDlg) that also force ISO formatting.

Personally I would prefer to go with option 3. What do you think ?

Meanwhile I will squash the commits and apply the others changes.

NathanRichard:
Squash and small corrections are done.

Note that on the previous PR, one of the remarks what that names should be camelCased. I applied it to this PR, with the exception of BaseDialog.GetPath(), as it must be able to pass as a WX Widget for the Chooser class and GetPath() is required for the duck typing to work.

I think the last point is the date formatting, please see my reply before.

Cheers,

chriskiehl:
Apologies for the delay! 

Your research seems sound. Thanks for diving into the weeds on this! Let's go with Option 3. 

Can you add a tiny summary of the behavior to your doc updates? Just a one liner explaining the ISO transform.

NathanRichard:
Hello,
No need to apologize, as you can see I am not very reactive myself. ;)
I added a comment about the ISO conversion in the doc and rebased upon the latest changes.
Is it good for integration ? :)
Best regards

chriskiehl:
Github is confused, but this was merged into 1.0.4 here: https://github.com/chriskiehl/Gooey/commits/1.0.4-release. Closing this one. 

Thanks for the work and research (and patience) on this! :+1: :+1:

kkinzler:
I'm interested in trying out the TimeChooser. Is there an easy way to add it to my Gooey package? I wasn't seeing that 1.0.4 was available yet. Thanks!

chriskiehl:
@kkinzler you can pip install git branches if you want to preview 1.0.4: 

Checkout this SO post: https://stackoverflow.com/questions/20101834/pip-install-from-git-repo-branch 

note: it may be a little unstable, but feel free to report any issues you run into

PR #514, opened 2019-12-18
Closes #200 substitute command line arguments for defaults

### LINKED ISSUE #200: Feature request: Substitute command-line arguments for Gooey defaults

jschultz:
This simple patch make arguments specified on the command line override defaults specified in the Python script. The only messy part is allowing required arguments to be omitted on the command line. This is done by iterating through the actions and mutex groups adjusting the `required` and `nargs` field as appropriate.

I have based this patch on the master rather than the 1.0.4 release branch because the latter appears broken, as described [here](https://github.com/chriskiehl/Gooey/commit/4140f8e8e69585d1a421757d1fc1944b6de6bf36#commitcomment-36476094).

Support for this functionality has been expressed [here](https://github.com/chriskiehl/Gooey/issues/200#issuecomment-534391679), [here](https://github.com/chriskiehl/Gooey/pull/203#issuecomment-379490966) and [here](https://github.com/chriskiehl/Gooey/pull/203#issuecomment-466977921)

### DIFF
--- a/README.md
+++ b/README.md
@@ -258,6 +258,7 @@ Just about everything in Gooey's overall look and feel can be customized by pass
 | show_restart_button | Toggles whether or not to show the restart button at the end of execution |
 | run_validators | Controls whether or not to have Gooey perform validation before calling your program |
 | poll_external_updates | (Experimental!) When True, Gooey will call your code with a `gooey-seed-ui` CLI argument and use the response to fill out dynamic values in the UI (See: [Using Dynamic Values](#using-dynamic-values))|
+| use_cmd_args | Substitute any command line arguments provided at run time for the default values specified in the Gooey configuration |
 | return_to_config | When True, Gooey will return to the configuration settings window upon successful run |
 | progress_regex | A text regex used to pattern match runtime progress information. See: [Showing Progress](#showing-progress) for a detailed how-to | 
 | progress_expr | A python expression applied to any matches found via the `progress_regex`. See: [Showing Progress](#showing-progress) for a detailed how-to |
--- a/gooey/python_bindings/cmd_args.py
+++ b/gooey/python_bindings/cmd_args.py
@@ -1,0 +1,79 @@
+'''
+Created on Jan 15 2019
+
+@author: Jonathan Schultz
+
+This file contains code that allows the default argument values to be specified
+on the command line.
+'''
+
+from argparse import _SubParsersAction
+
+def parse_cmd_args(self, args=None):
+
+  def prepare_to_read_cmd_args(item):
+    '''
+    Before reading the command-line arguments, we need to fudge a few things:
+      1. If there are subparsers, we need a dest in order to know in which
+         subparser the command-line values should be stored.
+      2. Any required argument or mutex group needs to be made not required,
+         otherwise it will be compulsory to enter those values on the command
+         line.
+    We save the everything as it was before the fudge, so we can restore later.
+    '''
+    for action in item._actions:
+      if isinstance(action, _SubParsersAction):
+        action.save_dest = action.dest
+        if not action.dest:
+          action.dest = '_subparser'
+      else:
+        action.save_required = action.required
+        action.required = False
+        action.save_nargs = action.nargs
+        if action.nargs == '+':
+          action.nargs = '*'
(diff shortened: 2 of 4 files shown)


jschultz:
Hi @chriskiehl,

Thanks for the review. :)

The high level tour of the code is pretty quick. In short, it calls ``argparse.parse_args`` aka ``self.original_parse_args`` with the supplied command-line arguments, and overwrites argparse's default values with the result. The only issue that arises is how to deal with validation errors on the command line. In particular, you don't want to enforce required arguments at this stage but rather just leave them empty and let Gooey enforce the requirement. Ideally, ``argparse`` would have some arrangement where you can tell it what to enforce (like ``parse_known_args`` doesn't enforce 'unknown argument'). But as a fallback that works right now, the code iterates through the arguments and replaces any that are required (ie ``required == True or nargs in ['+', None]``) with non-required equivalents, then calls ``argparse.parse_args``, and finally restores the original values.

When you say move it out of the decorator, do you just mean put it into a separate function? Otherwise I'm afraid  I think it actually needs to happen in the decorator. I did write an earlier version of this functionality that didn't do the work inside the decorator and it became much more messy and complicated. Overwriting the defaults before calling ``config_generator.create_from_parser`` seems the neatest solution.

Excuse me for missing something, but what boundaries can you see that need enforcing?

Tests and documentation... OK yes I'll figure something out there.

And finally, enabling via a flag. I'm not opposed to what you say but could I first explain why I chose to make it default behaviour? As it stands, if you call a Gooey-enabled script and provide command-line arguments, those are silently ignored. It seems no worse to me that they be silently used to fill in values before the Gooey window appears, If that behaviour is to be an option, then consistency would suggest that when the option is disabled, the script should return an error code if it is called with arguments (other than ``--ignore-gooey``). What do you think?

chriskiehl:
>then calls argparse.parse_args, and finally restores the original values. 

Ohhhhh, I understand now. Yeah, this could definitely use a few comments here and there to guide through the intuitive parts. I often forget what even goes on in parts of Gooey any more, so anything that annotates any of the weird mutations that go on is valuable. 

>When you say move it out of the decorator, do you just mean put it into a separate function?  

Exactly. [Single Responsibility](http://principles-wiki.net/principles:single_responsibility_principle?redirect=1) and what not. I'd pull it out into a separate file so that it can be laser focused on doing just this tiny piece of functionality you're trying to enable. Then, yeah, you'd delegate to your code from the decorator. Having it all inline is just messy. 

>enabling via a flag. 

The summary of my position is that I'm _extremely_ hesitant to change the behavior of software -- especially software that has, at this point, been around awhile. I think we're too cavalier with behavioral changes as an industry. Once something's out in the wild, you have no idea how people will actually use it, which quirks they're used to, how they've worked around limitations, etc.. I've failed from time to time (actually repeatedly), but I generally (and to great maintenance pain) try to make accretions to Gooey's functionality non-breaking and backwards compatible. So, while everyone generally views their perception of software / feature / use case to be the intuitive one or the way things should be done, I tend to err on the side of caution. 

So, the short version is, your argument is perfectly reasonable, but I ultimately disagree with it (hopefully, this is perceived as friendly as possible!) ^_^ opt-in is the only way I'd be comfortable due to legacy reasons.

jschultz:
OK no problems. As a package maintainer I can understand the imperative to not change existing behaviour.

Of course I found some problems in doing the testing - it didn't handle subparsers properly. So that is fixed now too.

I think it should all be good now. Hope you don't mind the separate commits - maybe I should merge them?

chriskiehl:
lgtm. Nice cleanup! One last thing, could you swap this to point at the release branch (1.0.4-release)?

chriskiehl:
Oh, also, yeah, please squash your commits down into one.

Issue #523, opened 2020-01-12
Add ColorChooser widget


conradhilley:
Add a `ColorChooser` widget option via `wx.ColourDialog`. 

Example below where hex is returned from the chooser and the input dialogs are colored such that their background and text colors match the selected color.

![image](https://user-images.githubusercontent.com/21027844/72227206-91ea6400-354e-11ea-90d5-d709a13fd0f8.png)

![image](https://user-images.githubusercontent.com/21027844/72227182-4b950500-354e-11ea-9983-056fd8a5f976.png)

chriskiehl:
Merged. :+1:

PR #524, opened 2020-01-19
closes 523 - add colorchooser widget

conradhilley:
Closes https://github.com/chriskiehl/Gooey/issues/523, adds ColourChooser widget which returns hex color codes. I went with the spelling `Colour` in order to keep with wx conventions.

I can add an example to https://github.com/chriskiehl/GooeyExamples, once approved.

colour_chooser_example


chriskiehl:
This is just delightful. 

Merged!

PR #525, opened 2020-02-10
Reimplemented output printing as per PR#502

tom-p-reichel:
Moved logic from old pull request into the menubar system, had to change the constructor for menubar because it did not have a reference to the application-- and could not otherwise access the console text. Added corresponding docs in root README and a test file. New logic since last pull allows user to specify font family and size when printing, optionally.


README.md
```diff
@@ -429,11 +429,12 @@ Individual menu items in a group are also just maps of key / value pairs. Their
 * `menuTitle` - the name for this MenuItem  
 
 
-Currently, three types of menu options are supported: 
+Currently, four types of menu options are supported: 
 
  * AboutDialog 
  * MessageDialog
  * Link
+ * PrintDialog
```
chriskiehl: Awesome! Thanks for updating the docs!

gooey/gui/components/menubar.py
```diff
@@ -78,4 +80,23 @@ def spawnAboutDialog(self, item, *args, **kwargs):
...
+            "MODERN" : wx.FONTFAMILY_MODERN,
+            "TELETYPE" : wx.FONTFAMILY_TELETYPE,
+            "MAX" : wx.FONTFAMILY_MAX
+        }
+        printer = wx.richtext.RichTextPrinting()
+        buf = wx.richtext.RichTextBuffer()
+        buf.BeginFont(wx.Font(item.get("size",10),
+            fonts[item.get("font","TELETYPE")],
+            wx.FONTSTYLE_NORMAL,wx.FONTWEIGHT_NORMAL))
+        buf.AddParagraph(self.application.console.getText())
```
chriskiehl: (thinking out loud here) I wonder if there's a better approach to this. As is, it feels weird that a Menu Bar would know, and have concrete reference to, the main application (_and_ drill down into the _application's_ dependency to get the console text). It's a bit Demeter-y smell wise. 

Maybe the dependencies should go the other way, so the `application` holds a dependency on a `PrintManager` kind of class that houses all the "how to" related to Printing, and then all this Menu fella would be responsible for was dispatching an event via the PubSub mechanisms?

gooey/gui/components/menubar.py
```diff
@@ -78,4 +80,23 @@ def spawnAboutDialog(self, item, *args, **kwargs):
             if field in item:
                 getattr(about, method)(item[field])
 
-        three_to_four.AboutBox(about)
\ No newline at end of file
+        three_to_four.AboutBox(about)
+
+    def spawnPrintDialog(self, item, *args, **kwargs):
+        fonts = {
```
chriskiehl: I lean towards just using whatever is set via the `terminal_font_xxx` family of options (see [here](https://github.com/chriskiehl/Gooey/blob/master/gooey/python_bindings/config_generator.py#L90)), rather than introducing another set of font options just for printing. I'd wager that whatever font they choose for the terminal, will be the one they want to see printed out. 

Lemme know if you have an alternative use case in mind that warrants the extra font controls, though.

chriskiehl:
Closing due to Inactivity

PR #538, opened 2020-03-26
[Closes #537] Return Action when calling add_argument

### LINKED ISSUE #537: add_argument returns None instead of argparse.Action
When adding an argument to a parser with add_argument, argparse will return the Action associated with the argument. This functionality is lost in GooeyParser's add_argument function.

While not well documented in the Python docs, this is a powerful feature of Argparse and not having it blocks (or at least makes more difficult than necessary) some in-depth use of custom Actions with Gooey.

donaldst:
A simple change that should enable certain patterns like grouping Actions, triggering Actions on the fly, etc. when using a GooeyParser rather than an ArgumentParser.


chriskiehl:
lgtm. Nice fix! 👍

Issue #550, opened 2020-04-25
MultiDirChooser fails for paths with spaces


besslwalker:
- [x] OS - Windows 7
 - [x] Python Version - 3.8
 - [x] Gooey Version - 1.0.3 (also confirmed in 1.0.4 release branch)
 - [x] Thorough description of problem 
            When opening a directory with spaces in the path from the MultiDirChooser, I expect to receive the entire path as one item in the argument's list, e.g:

            
            Dirnames received:
            whoops/got some spaces here
            good_spaceless_name
     - [x] Actual Behavior 
           I receive one item in the argument's list for each part of the path that can be space-separated, e.g.:

           Dirnames received:
           whoops/got
           some
           spaces
           here
           good_spaceless_name
 - [x] A minimal code example -- preferably copy/pastable in the issue itself (less time figuring out how to run your code == more time debugging!) 
(Select at least one directory with spaces in the path, observe debugging output.)

```
from gooey import Gooey
from gooey import GooeyParser

@Gooey
def main(args):
    parser = GooeyParser()
    parser.add_argument("dirnames", nargs = "+", widget="MultiDirChooser")
    args = parser.parse_args(args)

    print("Dirnames recieved:")
    for d in args.dirnames:
        print(d)

if __name__ == '__main__':
    import sys
    main(sys.argv[1:])
```

 - [x] Screenshot (if visual quirk) N/A
 - [x] Anything else you may think will be helpful 

The fix is almost certainly the same or very similar to #544. I intend to have a pull request for you shortly.

chriskiehl:
Pull request now merged! Thanks for the fix!

PR #551, opened 2020-04-25
Closes Issue #550

besslwalker:
Issue #550: MultiDirChooser cannot handle paths with spaces.

Fix: instead of the general formatter, use formatters.multiFileChooser, which handles spaces in paths without problems.


besslwalker:
Whoops, branched from 1.0.4 after my other change. I'll get you a cleaner one.

Issue #549, opened 2020-04-25
possible issue with wxpython April24 update?


bradh11:
- [windows 10 ] OS
 - [ 3.8.2] Python Version 
 - [1.0.3 ] Gooey Version 
 - [ It seems i have a bug in code that was working until recently.  The error happens upon launch.  Code runs perfectly without the @Gooey decorator] Thorough description of problem 
     - [ launch gui for my argparse app] Expected Behavior 
     - [ stacktrace] Actual Behavior 
 - [see below for error ] A minimal code example -- preferably copy/pastable in the issue itself (less time figuring out how to run your code == more time debugging!) 
 - [n/a - gui doesnt launch ] Screenshot (if visual quirk) 
 - [ I think this is related to yesterdays wxpython 4.1.0 update...] Anything else you may think will be helpful 

 ```
$ python cucm-exporter.py 
status: starting 2020-04-25 08:41:26.255955
Traceback (most recent call last):
  File "cucm-exporter.py", line 287, in <module>
    main()
  File "C:\Users\bradh\Documents\code\cucm-exporter\.venv\lib\site-packages\gooey\python_bindings\gooey_decorator.py", line 87, in inner2
    return payload(*args, **kwargs)
  File "cucm-exporter.py", line 214, in main
    cli_args = parser.parse_args()
  File "C:\Users\bradh\Documents\code\cucm-exporter\.venv\lib\site-packages\gooey\python_bindings\gooey_parser.py", line 114, in parse_args
    return self.parser.parse_args(args, namespace)
  File "C:\Users\bradh\Documents\code\cucm-exporter\.venv\lib\site-packages\gooey\python_bindings\gooey_decorator.py", line 82, in run_gooey
    application.run(build_spec)
  File "C:\Users\bradh\Documents\code\cucm-exporter\.venv\lib\site-packages\gooey\gui\application.py", line 21, in run
    app = build_app(build_spec)
  File "C:\Users\bradh\Documents\code\cucm-exporter\.venv\lib\site-packages\gooey\gui\application.py", line 30, in build_app
    gapp = GooeyApplication(merge(build_spec, imagesPaths))
  File "C:\Users\bradh\Documents\code\cucm-exporter\.venv\lib\site-packages\gooey\gui\containers\application.py", line 42, in __init__
    self.header = FrameHeader(self, buildSpec)
[...]

PR #555, opened 2020-04-26
Update the wxPython version

KrzysiekJ:
Submitting this pull request to the 1.0.3.1-hotfix branch because it seems most natural in this case.

Apparently the previous version of wxPython was causing an undocumented problem with “source code string cannot contain null bytes”. Steps to reproduce:

```
pex 'wxpython==4.0.7' --platform win_amd64-cp-38-cp38 -o example.pyz
```

Then (on Windows) `py example.pyz` and `import wx`.

If we replace `'wxpython==4.0.7'` with `'wxpython==4.0.7.post2'`, it starts to work (don’t know if it is actually related to PEX or not).

### DIFF
--- a/gooey/__init__.py
+++ b/gooey/__init__.py
@@ -2,4 +2,4 @@ import os
 from gooey.python_bindings.gooey_decorator import Gooey
 from gooey.python_bindings.gooey_parser import GooeyParser
 from gooey.gui.util.freeze import localResourcePath as local_resource_path
-__version__ = '1.0.3.1'
+__version__ = '1.0.3.2'
--- a/setup.py
+++ b/setup.py
@@ -6,7 +6,7 @@ from setuptools import setup, find_packages
 with open('README.md') as readme:
     long_description = readme.read()
 
-version = '1.0.3.1'
+version = '1.0.3.2'
 
 deps = [
     'Pillow>=4.3.0',
@@ -15,7 +15,7 @@ deps = [
 ]
 
 if sys.version[0] == '3':
-    deps.append('wxpython==4.0.7')
+    deps.append('wxpython==4.0.7.post2')
 
 
 setup(


SamuelMarks:
FYI: 4.1.0 is out now

GadgetSteve:
In general the lowest maintenance burden for  both this project and also any that interact with it would be to have:
wxpython >= 4.0.0, <5.0 _assuming that a quick test shows that it works with 4.0_ and only change the 5.0 to a lower release once one has been found that is incompatible if one is. 
This will save version thrashing with minimal risk. As wxpython uses semantic versioning we can safely expect that a major version change will be a breaking one but that until that change comes along there **probably** will not be problems.

chriskiehl:
@GadgetSteve  

>uses semantic versioning we can safely expect 

Ho boy, I cannot describe to you how little I expect semantic versioning to _mean_ anything in practice ^_^ 

My lack of faith aside, in the specific wx case, the versioning scheme is not semver and instead a custom even / odd setup for which versions are stable. Sitting on a `>= 4.0` style policy has already [bit the project once in the past](https://github.com/chriskiehl/Gooey/issues/549). So, to avoid breaking people who use Gooey due to a, by design, unstable minor wx version bump, Gooey will continue to periodically onboard new wx versions after a period of stability / testing.

bilderbuchi:
Is there a possibility to relax the version specified to say `wxpython==4.0` instead of `4.0.7`? This should still keep you on the "stable" even-numbered minors and avoid API breakage (and you can bump it to 4.2 in the future) but will still get bug fixes. In addition, it avoids weird issues like [this](https://github.com/conda-forge/gooey-feedstock/commit/166521596ab14545f6e886df8204b0952b26943a) triggered by 4.0.7 (without .post2) not being available everywhere.

I am just weird conflict dependency issues in one of my packages (only at runtime, though!), and I suspect the strict depencency somehow being the culprit (but can't confirm that now).

PR #557, opened 2020-05-11
Add instructions for conda install

jfcorbett:
Hi, I made Gooey available on conda forge ([recipe](https://github.com/conda-forge/gooey-feedstock))

I thought it would be useful to let your users know that it can be installed this way as well as with pip.

Let me know if you would like to be co-maintainer of the conda recipe. Right now it's just me.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,10 @@ Table of Contents
 The easiest way to install Gooey is via `pip`
 
     pip install Gooey 
+	
+or `conda`
+
+    conda install Gooey -c conda-forge
 
 Alternatively, you can install Gooey by cloning the project to your local directory
 


chriskiehl:
Howdy!

This is neat. I think this would be a great candidate for the [Wiki](https://github.com/chriskiehl/Gooey/wiki). Do you want to add it there as an alternative installation option? 

From a maintenance perspective, I can't put it in the main README as an officially endorsed installation path. I don't know enough about Conda to support it. Long after you've moved on to other things, I'll still be here, so I've gotta be careful about what I can/cannot support.

jfcorbett:
Fair enough. I added a note in the wiki.

Issue #296, opened 2018-06-12
Fallback to CLI?


gsemet:
Is there a way to fallback to an argparse-like command line interface when the current python is not framework aware?
Or if the gooey program runs inside a CI? I would like to write a CLI command line tool with Gooey interface on top of it

gsemet:
All it take would be:
- give user the ability to change the "IGNORE_COMMAND" (`--ignore-gooey`) parameter to disable gooey. I would prefer to have a `--gui` option to force the display of gooey, without it:
  - if `--help`, display the help message in the console
  - if all commands are fulled, execute the program without building gooey
  - if a parameter is missing, display the GUI, and fallback to default CLI behavior in case of wx/python framework issue.

gsemet:
Here is my current solution:
```python


class AutoGooeyExecutor():

    def __init__(self):
        self.auto = True
        self.show_gui = True
        if "--gui" in sys.argv:
            self.auto = False
            self.show_gui = True
        if "--help" in sys.argv or "-h" in sys.argv:
            self.auto = False
            self.show_gui = False

    def execute(self):
        if self.auto:
            print("DBG: auto GUI")
            try:

                @Gooey(dump_build_config=False, program_name="Widget Demo", terminal_font_x=10)
                def gooey_execute():
                    return self.payload()

                return gooey_execute()
            except:
                print("DBG: fallback no GUI")
                return self.payload()
        elif self.show_gui:
            print("DBG: force GUI")

            @Gooey(dump_build_config=False, program_name="Widget Demo", terminal_font_x=10)
            def gooey_execute():
                return self.payload()

            return gooey_execute()
        else:
            print("DBG: fallback no GUI")
            return self.payload()

    def payload(self):
        desc = "Example application to show Gooey's various widgets"
        file_help_msg = "Name of the file you want to process"

        my_cool_parser = GooeyParser(description=desc)
        ....
        return 0

def main():
    age = AutoGooeyExecutor()
    return age.execute()
```

AlexanderWingard:
I also wanted something like this. To have Gooey hide under a subcommand and keep the argparse application working like it was. Here is my workaround:
```python
import argparse
from gooey import Gooey

gui = False;
def main():
    global gui
    parser = argparse.ArgumentParser()
    parser.add_argument('--ignore-gooey', action='store_true', help=argparse.SUPPRESS)

    subparsers = parser.add_subparsers(help='sub-command help', dest="sp")
    if not gui:
[...]

PR #568, opened 2020-05-25
Behave like --ignore-gooey has been passed if the display is missing

bje-:
This patch makes Gooey behave like --ignore-gooey has been passed if the display is missing. This allows Gooey applications to run unmodified on systems with no X display present.

Partially solves #296.


bje-:
Note that this is not necessarily a polished patch -- just a proof of concept!

PR #572, opened 2020-05-28
Silence Python 3.8 runtime warning

bje-:
This patch removes the buffer size paraemter in calls to `subprocess.Popen()` so that the default value is assumed. This silences this Python 3.8 runtime warning:

```
RuntimeWarning: line buffering (buffering=1) isn't supported in binary mode, the default buffer size will be used```


chriskiehl:
Checkout [CONTRIBUTING.md](https://github.com/chriskiehl/Gooey/blob/master/CONTRIBUTING.md ). These should target the release branch.

Issue #569, opened 2020-05-25
Gooey crashes on count actions when default=0


bje-:
- [Ubuntu 18.04 ] OS
 - [ 3.6.9] Python Version 
 - [ 1.0.3.1] Gooey Version 
 - Description of the problem: When using `argparse` count actions with a default value of zero, Gooey crashes on startup (backtrace below). I was able to workaround this by not setting a default, but then the default is `None` (undesirable). This usage comes straight from the Python `argparse` documentation.
     - Expected Behavior: No crash
     - Actual Behavior: Backtrace follows
```
Traceback (most recent call last):
  File "foo.py", line 10, in <module>
    args = process_options()
  File "/home/bje/.local/lib/python3.6/site-packages/gooey/python_bindings/gooey_decorator.py", line 87, in inner2
    return payload(*args, **kwargs)
  File "foo.py", line 8, in process_options
    return parser.parse_args()
  File "/home/bje/.local/lib/python3.6/site-packages/gooey/python_bindings/gooey_decorator.py", line 82, in run_gooey
    application.run(build_spec)
  File "/home/bje/.local/lib/python3.6/site-packages/gooey/gui/application.py", line 21, in run
    app = build_app(build_spec)
  File "/home/bje/.local/lib/python3.6/site-packages/gooey/gui/application.py", line 30, in build_app
    gapp = GooeyApplication(merge(build_spec, imagesPaths))
  File "/home/bje/.local/lib/python3.6/site-packages/gooey/gui/containers/application.py", line 43, in __init__
    self.configs = self.buildConfigPanels(self)
  File "/home/bje/.local/lib/python3.6/site-packages/gooey/gui/containers/application.py", line 221, in buildConfigPanels
    for widgets in self.buildSpec['widgets'].values()]
  File "/home/bje/.local/lib/python3.6/site-packages/gooey/gui/containers/application.py", line 221, in <listcomp>
    for widgets in self.buildSpec['widgets'].values()]
  File "/home/bje/.local/lib/python3.6/site-packages/gooey/gui/components/config.py", line 16, in __init__
    self.layoutComponent()
  File "/home/bje/.local/lib/python3.6/site-packages/gooey/gui/components/config.py", line 92, in layoutComponent
    self.makeGroup(self, sizer, item, 0, wx.EXPAND | wx.ALL, 10)
  File "/home/bje/.local/lib/python3.6/site-packages/gooey/gui/components/config.py", line 136, in makeGroup
[...]

PR #575, opened 2020-06-03
Pre-fill counter dropdowns with 0 (not 1) to 10 inclusive 

bje-:
Fixes issue #569.


chriskiehl:
:+1:

bje-:
Is there a schedule in mind for the 1.0.4 release that would include this bug fix?

chriskiehl:
The answer boils down to when I can carve out the time. 

While I would love to devote my full energy to Gooey, open source does not pay bills (or anything at all!). I assure you nobody is more bothered by its slow release cycle than me.

PR #579, opened 2020-06-03
Behave like --ignore-gooey has been passed if the display is missing

bje-:
This patch makes Gooey behave like --ignore-gooey has been passed if the display is missing. This allows Gooey applications to run unmodified on systems with no X display present.

Partially solves #296.

### DIFF
--- a/gooey/python_bindings/gooey_decorator.py
+++ b/gooey/python_bindings/gooey_decorator.py
@@ -10,6 +10,7 @@ import json
 import os
 import sys
 from argparse import ArgumentParser
+import wx
 
 from gooey.gui.util.freeze import getResourcePath
 from gooey.util.functional import merge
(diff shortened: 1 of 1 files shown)


bje-:
Up until now, `gooey_decorator.py` has not needed to import `wx`. Is this necessarily a problem?

bje-:
Another option that may be more acceptable (but could be documented) is for Gooey users to wrap the @Gooey decorator with a custom decorator that returns the original function if there is no display present, eg:

```
def conditionalGooey(*pargs, **kwargs):
    def decorator(func):
        if not wx.PyApp.IsDisplayAvailable():
            return func
        return Gooey(*pargs, **kwargs)(func)
    return decorator
```

chriskiehl:
I think your second comment is a good approach to the problem you're trying to solve. This isn't something I foresee Gooey dealing with directly. 

Your `conditionalGooey` would make for a nice cookbook wiki entry (hint, hint ^_^)

PR #589, opened 2020-06-27
Elapsed / Remaining Time Next to Progress Bar, closes #587 

### LINKED ISSUE #587: Show Time Remaining when progress_bar utilised
Hello!

I absolutely love this package, thank you so much!

One thing that I've found myself doing across projects using gooey is making an interface between Gooey and [tqdm](https://github.com/tqdm/tqdm).

So I decided I'd make it inbuilt into Gooey! I've already created the feature along with tests and such, so this issue is a formality to ensure compliance with the contributing guidelines. Find the branch at: <https://github.com/JackMcKew/Gooey/tree/1.0.4-release-tqdm-loading-bar>, I will create a PR just after this issue is submitted.

> The implementation is a Python 2.7 and 3 compatible 'extract' of the functionality in TQDM for Gooey

Furthemore, I've created a working example to add to the examples repository as well!

fbHcpCAGD8

JackMcKew:
Hello!

This the pull request to go with recently raised <https://github.com/chriskiehl/Gooey/issues/587>.

Changes are as follows:

- New pub/sub event created for application in `events.py` for `TIME_REMAINING_UPDATE`.
- StaticText label added in Footer adjacent to progress bar for displaying elapsed & remaining time.
- Functions added to `processor.py`:
    - `_get_current_time()` either use `time.perf_counter()` in Python 3.3+ or `timeit.default_timer()` for Python 2.7 compatibility to get a current time in seconds as a float
    - `_calculate_time_remaining(progress,start_time)` estimate the time remaining by calculating the current rate of progress over elapsed time and extrapolate for the remainder of progress.
    - `format_interval()` This was taken from the tqdm project <https://github.com/tqdm/tqdm/blob/0cd9448b2bc08125e74538a2aea6af42ee1a7b6f/tqdm/std.py#L228> for showing the elapsed & remaining time in [H]:MM:SS format.
- `_forward_stdout` in `processor.py` has been altered to show time remaining when progress is greater than 0.
- Relevant tests for all functions added can be found in `test_processor.py`.

I absolutely love this package, thank you so much!

> This does not include a dependency of tqdm, as the functionality has been 'extracted'.

Furthemore, I've created a working example to add to the examples repository as well! Which I have created a PR over at: <https://github.com/chriskiehl/GooeyExamples/pull/31>.

fbHcpCAGD8

Round 2 at pull request!

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,32 +18,33 @@ Table of Contents
 -----------------  
 
 - [Gooey](#gooey)
-- [Table of contents](#table-of-contents)
-- [Latest Update](#latest-update)
-- [Quick Start](#quick-start)
-    - [Installation Instructions](#installation-instructions)
+  - [Table of Contents](#table-of-contents)
+  - [Quick Start](#quick-start)
+    - [Installation instructions](#installation-instructions)
     - [Usage](#usage)
     - [Examples](#examples)
-- [What It Is](#what-is-it)
-- [Why Is It](#why)
-- [Who is this for](#who-is-this-for)
-- [How does it work](#how-does-it-work)
-- [Internationalization](#internationalization)
-- [Global Configuration](#global-configuration)
-- [Layout Customization](#layout-customization)
-- [Run Modes](#run-modes)
-    - [Full/Advanced](#advanced)
+  - [What is it?](#what-is-it)
+  - [Why?](#why)
+  - [Who is this for?](#who-is-this-for)
+  - [How does it work?](#how-does-it-work)
+      - [Mappings:](#mappings)
+    - [GooeyParser](#gooeyparser)
+  - [Internationalization](#internationalization)
+  - [Global Configuration](#global-configuration)
+  - [Layout Customization](#layout-customization)
+  - [Run Modes](#run-modes)
+    - [Advanced](#advanced)
     - [Basic](#basic)
     - [No Config](#no-config)
-- [Menus](#menus)    
-- [Input Validation](#input-validation)
-- [Using Dynamic Values](#using-dynamic-values)
-- [Showing Progress](#showing-progress)
-- [Customizing Icons](#customizing-icons)
-- [Packaging](#packaging)
-- [Screenshots](#screenshots)
-- [Contributing](#wanna-help)
-- [Image Credits](#image-credits)
+    - [Menus](#menus)
+    - [Input Validation](#input-validation)
+  - [Using Dynamic Values](#using-dynamic-values)
+  - [Showing Progress](#showing-progress)
+    - [Elapsed / Remaining Time](#elapsed--remaining-time)
+  - [Customizing Icons](#customizing-icons)
+  - [Packaging](#packaging)
+  - [Screenshots](#screenshots)
+  - [Wanna help?](#wanna-help)
 
 
 
@@ -156,16 +157,16 @@ At run-time, it parses your Python script for all references to `ArgumentParser`
 
 Gooey does its best to choose sensible defaults based on the options it finds. Currently, `ArgumentParser._actions` are mapped to the following `WX` components. 
 
-| Parser Action    | Widget    | Example |
-|:----------------------|-----------|------|
-| store  |  TextCtrl |  <img src="https://github.com/chriskiehl/GooeyImages/raw/images/readme-images/f54e9f5e-07c5-11e5-86e5-82f011c538cf.png"/>|
-| store_const   |     CheckBox |  <img src="https://github.com/chriskiehl/GooeyImages/raw/images/readme-images/f538c850-07c5-11e5-8cbe-864badfa54a9.png"/>|
-|   store_true|        CheckBox | <img src="https://github.com/chriskiehl/GooeyImages/raw/images/readme-images/f538c850-07c5-11e5-8cbe-864badfa54a9.png"/>|
-|  store_False  |      CheckBox|  <img src="https://github.com/chriskiehl/GooeyImages/raw/images/readme-images/f538c850-07c5-11e5-8cbe-864badfa54a9.png"/>   |
-|       append |       TextCtrl |  <img src="https://github.com/chriskiehl/GooeyImages/raw/images/readme-images/f54e9f5e-07c5-11e5-86e5-82f011c538cf.png"/>  | 
-|        count|              DropDown &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; | <img src="https://github.com/chriskiehl/GooeyImages/raw/images/readme-images/f53ccbe4-07c5-11e5-80e5-510e2aa22922.png"/> | 
-| Mutually Exclusive Group | RadioGroup | <img src="https://github.com/chriskiehl/GooeyImages/raw/images/readme-images/f553feb8-07c5-11e5-9d5b-eaa4772075a9.png"/>
-|choice &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;|        DropDown | <img src="https://github.com/chriskiehl/GooeyImages/raw/images/readme-images/f54e4da6-07c5-11e5-9e66-d8e6d7f18ac6.png"/> |
+| Parser Action                                                                                                                                                                                                                                                                   | Widget                                                                                                    | Example                                                                                                                  |
+| :------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ | --------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------ |
+| store                                                                                                                                                                                                                                                                           | TextCtrl                                                                                                  | <img src="https://github.com/chriskiehl/GooeyImages/raw/images/readme-images/f54e9f5e-07c5-11e5-86e5-82f011c538cf.png"/> |
+| store_const                                                                                                                                                                                                                                                                     | CheckBox                                                                                                  | <img src="https://github.com/chriskiehl/GooeyImages/raw/images/readme-images/f538c850-07c5-11e5-8cbe-864badfa54a9.png"/> |
+| store_true                                                                                                                                                                                                                                                                      | CheckBox                                                                                                  | <img src="https://github.com/chriskiehl/GooeyImages/raw/images/readme-images/f538c850-07c5-11e5-8cbe-864badfa54a9.png"/> |
+| store_False                                                                                                                                                                                                                                                                     | CheckBox                                                                                                  | <img src="https://github.com/chriskiehl/GooeyImages/raw/images/readme-images/f538c850-07c5-11e5-8cbe-864badfa54a9.png"/> |
+| append                                                                                                                                                                                                                                                                          | TextCtrl                                                                                                  | <img src="https://github.com/chriskiehl/GooeyImages/raw/images/readme-images/f54e9f5e-07c5-11e5-86e5-82f011c538cf.png"/> |
+| count                                                                                                                                                                                                                                                                           | DropDown &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; | <img src="https://github.com/chriskiehl/GooeyImages/raw/images/readme-images/f53ccbe4-07c5-11e5-80e5-510e2aa22922.png"/> |
+| Mutually Exclusive Group                                                                                                                                                                                                                                                        | RadioGroup                                                                                                | <img src="https://github.com/chriskiehl/GooeyImages/raw/images/readme-images/f553feb8-07c5-11e5-9d5b-eaa4772075a9.png"/> |
+| choice &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; | DropDown                                                                                                  | <img src="https://github.com/chriskiehl/GooeyImages/raw/images/readme-images/f54e4da6-07c5-11e5-9e66-d8e6d7f18ac6.png"/> |
 
 ### GooeyParser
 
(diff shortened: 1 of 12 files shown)


JackMcKew:
Furthermore the README has been updated, although the display of the image is dependent on the pull request over on the GooeyImages repo: <https://github.com/chriskiehl/GooeyImages/pull/1>

JackMcKew:
When running all tests, there is also a failure on `test_cmd_args.py` but I don't think this is stemming from these change.

``` bash
======================================================================
ERROR: test_required_not_enforced (tests.test_cmd_args.TextCommandLine)
----------------------------------------------------------------------
  File "c:\Users\jackm\Documents\GitHub\Gooey\gooey\tests\test_cmd_args.py", line 32, in test_required_not_enforced
    cmd_args.parse_cmd_args(parser)
  File "c:\Users\jackm\Documents\GitHub\Gooey\gooey\python_bindings\cmd_args.py", line 78, in parse_cmd_args
    overwrite_default_values(self, self.original_parse_args(args))
  File "c:\users\jackm\.pyenv\pyenv-win\versions\3.8.1-amd64\lib\argparse.py", line 1771, in parse_args
    self.error(msg % ' '.join(argv))
  File "c:\users\jackm\.pyenv\pyenv-win\versions\3.8.1-amd64\lib\argparse.py", line 2521, in error
    self.exit(2, _('%(prog)s: error: %(message)s\n') % args)
  File "c:\users\jackm\.pyenv\pyenv-win\versions\3.8.1-amd64\lib\argparse.py", line 2508, in exit
    _sys.exit(status)
SystemExit: 2
```

chriskiehl:
Awesomely detailed PR! Additions to the README, Images, and Examples folder are 👍 

Couple of high level things from trying this out: 

1. the time doesn't appear to be super accurate? is there a way to tighten up the functionality a bit? Running some experiments on my Windows box, it was effectively random and confusing at times. 
1. It flickers like a mad man while updating[0]! There's a `transactUI` context function floating around which may help this. The general WX idea behind avoiding this sort of thing is freezing/thawing during updates. Checkout [this blog post](https://www.blog.pythonlibrary.org/2014/11/11/wxpython-freeze-and-thaw/) for some background info. 
1. The time display isn't aligned with the other components. A quick `wx.ALIGN_CENTER_VERTICAL` may alleviate that one.
1. It appears to stick around on the success screen. The progress bar hides after completion, so it feels odd for other progress info to stick around (doubly so when it ends on funky results[1]).   
1. It explodes if you have mixed content in the output. progress_expr can often be `None`
1. Configuration is the name of the game for Gooey. Showing/computing the time should be opted-in via a decorator argument. 
1. Squash your commits into a single high level feature commit : )

[0] example
[1] image
[2] This causes your changes to fail inside Processor when it finds `progress_expr` is None. 

```python 
def compute_range(length):
    for i in range(length):
        import time
        print('some log about an expensive computation', str(time.time()))
        time.sleep(.2)
        print(f"progress: {i}/{length}")
```

chriskiehl:
btw, on which platform are you seeing that test failure? I'm unable to reproduce on Windows and mainline Gooey version. 🤔

JackMcKew:
Thank you so much for the prompt & thorough review! 👍 

All your comments have been fixed apart from 1, as estimating time remaining is an arcane art. I will squash the commits now.

1. At the moment it relies on `progress` (which updates the progress bar), so there's potential to use a more accurate figure than 1-100, I'll have to look into how we can retrieve more accurate figures around number of finished iterations and total iterations. As per TQDM's source code as well (it truly is an arcane art), this is also especially difficult around short loops like this example: 

``` python
# if unspecified, attempt to use rate = average 
# (we allow manual override since predicting time is an arcane art)
```
2. This has been fixed by setting `Footer.setDoubleBuffered(True)` as per this thread <https://stackoverflow.com/questions/5294274/wxpython-statictext-widget-flickering>.
3. Fixed as per your comment 💯.
4. I personally like to see how long my programs run for (as I'm big on set-and-forget applications). Default behaviour now hides on complete. 
5. Fixed by check for `_progress` not being `None`.
6. New decorator keywords introduced `hide_time_remaining` default False for displaying next to progress bar, `show_time_remaining_on_complete` default False for not hiding text on complete.

GIF has now been updated in https://github.com/chriskiehl/GooeyImages/pull/1

JackMcKew:
I've made an attempt to squash but unsure if it's worked properly.

In regard to failing test:

OS: Windows 10 Education
Version: 10.0.18363 Build 18363
Python Version: Python 3.8.1 (tags/v3.8.1:1b293b6, Dec 18 2019, 23:11:46) [MSC v.1916 64 bit (AMD64)] on win32

Let me know if you need any more information

JackMcKew:
Squash is now fixed! :)

README.md
```diff
@@ -265,6 +266,8 @@ Just about everything in Gooey's overall look and feel can be customized by pass
 | progress_expr | A python expression applied to any matches found via the `progress_regex`. See: [Showing Progress](#showing-progress) for a detailed how-to |
 | hide_progress_msg | Option to hide textual progress updates which match the `progress_regex`. See: [Showing Progress](#showing-progress) for a detailed how-to |
 | disable_progress_bar_animation | Disable the progress bar |
+| hide_time_remaining | Disable the time remaining text see [Elapsed / Remaining Time](#elapsed--remaining-time) |
```
chriskiehl, line 269: See: top level comments on these 

Also, these should be opt _in_ not out.

gooey/gui/containers/application.py
```diff
@@ -231,6 +234,7 @@ def showSettings(self):
         self.header.setSubtitle(self.buildSpec['program_description'])
         self.footer.showButtons('cancel_button', 'start_button')
         self.footer.progress_bar.Show(False)
+        self.footer.time_remaining_text.Show(False)
```
chriskiehl, line 237: I think you forget to explicitly turn in on for the console screen (or there may be some other bug floating around)? 

If you run a program twice, the timing info doesn't show after the first invocation. e.g.

1. Run [your example](https://github.com/chriskiehl/GooeyExamples/pull/31). 
2. note that the timing info is showing during run 
3. once complete: Click restart / edit + start 
4. No timing information is shown.

gooey/gui/processor.py
```diff
@@ -59,16 +61,59 @@ def run(self, command):
         t = Thread(target=self._forward_stdout, args=(self._process,))
         t.start()
 
+    def format_interval(self,time_value):
```
chriskiehl, line 64: `Processor` is now doing a bit too much imo (it shouldn't have dealt with the finer details of progress here either, to be fair). You should pull out the time stuff into a diff module. It'll make your tests a bit clearer, too, as you won't have to call through `CommandProcessor` just to test that something gets formatted correctly.

gooey/tests/test_processor.py
```diff
@@ -9,34 +9,62 @@ class TestProcessor(unittest.TestCase):
     def test_extract_progress(self):
```
chriskiehl, line 9: You'll also need to add an integration(ish) test file which verifies the UI does the right things for the various options available. There's some prior art [here](https://github.com/chriskiehl/Gooey/blob/master/gooey/tests/test_header.py). 

A lot of regressions in Gooey have tended to be on the UI side as it slowly drifts out of sync with the rest of the code.

JackMcKew:
Holy dooly, first of all thank you so much for the comprehensive review!

This will all take some time to work through, but I'll try to hit all the points. Let me know if anything else pops up as I am working through them.

JackMcKew:
Changes thus far (still some further changes to come before squashing)

Just wondering what your thoughts would be regarding the dictionary structure decorator arguments below, as this would be in a lot of future changes as well (I think). I strongly agree that the timing options should be inbuilt in a `progress_options` dictionary argument, but I believe that should be a separate pull request.

- Move to dictionary structure decorator arguments
    - This introduces a new problem when a user doesn't specify all the values and thus overwrites with an half-filled dictionary causing errors. Unsure how to handle this, current thoughts are to merge default dictionary with user-specified, with user-specified overwriting default. To do this with Python 2 support would be introducing a new function in the functional module: <https://stackoverflow.com/questions/38987/how-do-i-merge-two-dictionaries-in-a-single-expression-in-python>
- Text is explicitly shown if setting is True, this alleviates the problem of editing/restarting
- Time module created in gooey/gui/util containing all timing functionality

**Remaining**
- Provide options in timing options to enable/disable either elapsed or remaining
- Integration tests
- New tests for time module

gooey/gui/components/footer.py
```diff
@@ -4,6 +4,7 @@
 from gooey.gui import events
 from gooey.gui.lang.i18n import _
 from gooey.gui.pubsub import pub
+from gooey.gui.util.wx_util import transactUI
```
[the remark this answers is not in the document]
JackMcKew: Amended :)

gooey/gui/containers/application.py
JackMcKew: Amended in `showConsole`.

gooey/gui/processor.py
JackMcKew: Separated into new module gooey/gui/util/time.

gooey/tests/test_processor.py
JackMcKew: Added integration tests similar to that of the test_header.py. Unsure how to test the onComplete screen though, will look into that

JackMcKew:
Tests are now implemented, only configurable elapsed vs remaining left to go, will do this after we have a way forward on the dictionary merging plan

.vscode/settings.json
```diff
@@ -0,0 +1,12 @@
+{
```
chriskiehl, line 1: Whoops. Committed accidentally?

JackMcKew:
Thank you for your input.

I'll get into making the sane defaults approach. This may be the instigator for another pull request moving default behaviour to the dictionary styled arguments.

I've created a new function in utils/functional.py to do this for us.

``` python
def merge_two_dictionaries(x,y):
    """
    Merge 2 dictionaries with y taking overwriting x if a key collision is found

    This is mainly useful for maintaining the dictionary arguments to allow for more expressive & extensible arguments.
    https://stackoverflow.com/questions/38987/how-do-i-merge-two-dictionaries-in-a-single-expression-in-python-taking-union-o

    Args:
        x (dict): Input dictionary
        y (dict): Input dictionary

    Returns:
        The combined dictionary of x & y with y taking preference on the occasion of key collision
    """

    try:
        return {**x,**y}
    except:
        z = x.copy()
        z.update(y)
        return z
```

JackMcKew:
@chriskiehl I'm trying to write the tests for this now that the merge functionality is there, although the time remaining text isn't show until the user hits start and there is some progress to go off of. Any thoughts on how to approach this programmatically?

JackMcKew:
Amended to have behaviour with a default dictionary which can be overriden by user. Unsure how to handle tests that depend on progress running to verify text is displayed or not. Support for Python 2 is included in the dictionary merging functionality.

Fully squashed, docs updated in README and Example updated over at https://github.com/chriskiehl/GooeyExamples/pull/31

gooey/gui/processor.py
```diff
@@ -64,11 +66,21 @@ def _forward_stdout(self, process):
         Reads the stdout of `process` and forwards lines and progress
         to any interested subscribers
         '''
+        _start_time = get_current_time()
```
chriskiehl, line 69: Arg, so unfortunately, it didn't occur to me the first time I looked at this, but there's kind of a core design issue with this which is a show stopper in its current implementation. 

It's dependent on the assumption that there will be steady regular output being written to stdout. However if you run your code using something which produces chunked / irregular output (due to network latency or just general slow work or whatever), the elapsed time ceases to give steady updates. 

You can simulate the problem with a slight modification to your [example](https://github.com/chriskiehl/GooeyExamples/pull/31). 

```python
def compute_range(length):
    for i in range(10):
        import time
        from random import randint
        time.sleep(randint(1,3))
        print(f"progress: {i}/{10}")
```

This will cause the elapse time output to hop randomly between 1-3 seconds rather than giving a smooth continuous output. 

So, for this to work in the general case, you'll need to break apart the production of a timing event from the publishing of activity related to stdout. 

One way to accomplish this would be:

1. Introduce a stand alone timing class which deals with publishing updates about the current run time.
    * this class will have an internal blob of data which houses the timing info (start, estimatedRemaining) 
    * it'll have a subscription on `PROGRESS_UPDATE` which it can use update its internal estimation 
2. This class will publish its timing info on some regular cadence via [wx.Timer](https://wxpython.org/Phoenix/docs/html/wx.Timer.html#wx.Timer) 

As a super-duper rough sketch, the idea would be something like... 


```python
class Timing(object):
    def __init__(self, parent):
        self.startTime = 0
        self.estimatedRemaining = None
        self.wxTimer = wx.Timer(parent)
        parent.Bind(wx.EVT_TIMER, self.publishTime, self.wxTimer)

        pub.subscribe(events.PROGRESS_UPDATE, self._updateEstimate)

    def _updateEstimate(self, *args, **kwargs):
        prog = kwargs.get('progress')
        # your computation here
        self.estimatedRemaining = 0

    def publishTime(self, *args, **kwargs):
        pub.send_message(
            events.TIME_UPDATED,
            start=self.startTime,
            current=time.time(),
            elapsed=time.time() - self.startTime,
            estimatedRemaining=self.estimatedRemaining)

    def start(self):
        self.startTime = time.time()
        self.estimatedRemaining = None
        self.wxTimer.Start(500)

    def stop(self):
        self.wxTimer.Stop()
```

This would allow timing events to be published regardless of whether or not there was activity coming over the hosted program's stdout.

gooey/gui/util/functional.py
```diff
@@ -1,6 +1,30 @@
+'''
+Utils for functional methodologies throughout Gooey
 
+'''
 
 
+def merge_two_dictionaries(x,y):
```
chriskiehl, line 7: I think just `merge` is ok as a name. The "two" part can be inferred from the number of arguments ^_^

gooey/python_bindings/config_generator.py
```diff
@@ -69,8 +71,11 @@ def create_from_parser(parser, source_path, **kwargs):
       'progress_regex':       kwargs.get('progress_regex'),
       'progress_expr':        kwargs.get('progress_expr'),
       'hide_progress_msg':    kwargs.get('hide_progress_msg', False),
+      'timing_options':       merge_two_dictionaries(gooey_decorator.defaults.get('timing_options'),kwargs.get('timing_options')),
       'disable_progress_bar_animation': kwargs.get('disable_progress_bar_animation'),
       'disable_stop_button':  kwargs.get('disable_stop_button'),
+      # 'hide_time_remaining':  kwargs.get('hide_time_remaining', False),
```
chriskiehl, line 77: Leftover comments?

gooey/tests/test_time_remaining.py
```diff
@@ -0,0 +1,47 @@
...
+from tests.harness import instrumentGooey
+
+class TestFooterTimeRemaining(unittest.TestCase):
+
+    def make_parser(self):
+        parser = ArgumentParser(description='description')
+        return parser
+
+
+    # TODO: This test needs to have a progress loop to show the time remaining
```
chriskiehl, line 14: The `instrumentGooey` thing'll spit out the instantiated Gooey object, this will be built from whatever parser is provided, so you can use varying parser setups to test Gooey does the Right Thing. 

So, for example, if you wanted to test the presentational stuff is working as expected (and this is usually where a lot of Gooey regressions occur), you could put Gooey into the console mode (`gooeyApp.showConsole()`) and then assert that your timing info is visible/hidden. 

Similarly, you can test that your updates work by dispatching an event and then checking the results.

gooey/gui/processor.py
```diff
@@ -64,11 +66,21 @@ def _forward_stdout(self, process):
...
         '''
+        _start_time = get_current_time()
         while True:
             line = process.stdout.readline()
             if not line:
                 break
             _progress = self._extract_progress(line)
+
+            # For displaying elapsed & estimated remaining time
+            if _progress is not None and _progress > 0 and self.show_time_remaining:
```
chriskiehl, line 77: See other comments. We should rip this stuff out.

gooey/gui/processor.py
JackMcKew: Finally coming back around to looking at this, I'll have a go at we'll see how far we get abstracting it away from progress

gooey/python_bindings/config_generator.py
JackMcKew: Resolved 😄

gooey/gui/processor.py
JackMcKew: Resolved 😄

gooey/gui/util/functional.py
JackMcKew: Resolved 😄

gooey/gui/processor.py
JackMcKew: Resolved 😄  This is now fixed, I'm updating the tests for it and then I'll squash it all and push

gooey/tests/test_time_remaining.py
JackMcKew: Tests are now added and passing 👍

JackMcKew:
Okay, the timing module has been abstracted into it's own using `wx.Timer` and the text utilizes pub/sub events to update as required.

Tests have been implemented for both and all passing.

Let me know if it doesn't behave on your machine as expected, there was a few artifacts left over from trying to squash and already squashed 6 times commits but I think I got there.

Hoping that this is getting closer to closing this out! 😄

gooey/gui/processor.py
chriskiehl: Apologies on my end for the delay! Glad you're back at it! Could you bring this up to date with 1.0.5? I wanna give it a whirl locally but it currently conflicts with the current state of things.

gooey/gui/util/time.py
```diff
@@ -0,0 +1,93 @@
...
+class Timing(object):
+
+    def __init__(self, parent):
+        self.startTime = 0
+        self.estimatedRemaining = None
+        self.wxTimer = wx.Timer(parent)
+        self.parent = parent
+        parent.Bind(wx.EVT_TIMER, self.publishTime, self.wxTimer)
+
+        pub.subscribe(events.PROGRESS_UPDATE, self._updateEstimate)
```
chriskiehl, line 17: There's a tiny part of me that wants to squabble over this line and as I endlessly bike shed over what goes where, but... I'll spare you my neurotic tendencies ^_^ Changes look good to me. Branch just needs to be brought up to date with current release, then I'll do some final sanity testing on my end, then should be good to go! 

Thanks for all your work on this!

JackMcKew: I've rebased on the `1.0.5-release` branch, and it's telling me there will be 71 pushes and 38 pulls, unsure if this will become apart of this pull request, and I've changed the target branch to merge into as `1.0.5-release`, let me know if this isn't correct

Furthermore, before I push all these changes onto this, I ran the tests (and tried to run a Gooey application) and I'm getting 65 failures and 1 error (and the application won't run), which all the errors are `wx._core.wxAssertionError: C++ assertion "Assert failure" failed at ..\..\src\msw\font.cpp(586) in wxNativeFontInfo::SetWeight(): unknown font weight`. This seems to be stemming from the console text object, so unsure if I need to alter anything with these changes?

```
  File "c:\Users\jackm\Documents\GitHub\Gooey\gooey\tests\test_time_remaining.py", line 31, in test_time_remaining_visibility_on_complete
    with instrumentGooey(self.make_parser(), timing_options=testdata) as (app, gooeyApp):
  File "c:\users\jackm\.pyenv\pyenv-win\versions\3.8.1-amd64\lib\contextlib.py", line 113, in __enter__
    return next(self.gen)
  File "c:\Users\jackm\Documents\GitHub\Gooey\gooey\tests\harness.py", line 18, in instrumentGooey
    app, gooey = application.build_app(buildspec)
  File "c:\Users\jackm\Documents\GitHub\Gooey\gooey\gui\application.py", line 32, in build_app
    gapp = GooeyApplication(merge(build_spec, imagesPaths))
  File "c:\Users\jackm\Documents\GitHub\Gooey\gooey\gui\containers\application.py", line 47, in __init__
    self.console = Console(self, buildSpec)
  File "c:\Users\jackm\Documents\GitHub\Gooey\gooey\gui\components\console.py", line 27, in __init__
    self.textbox.SetFont(wx.Font(
wx._core.wxAssertionError: C++ assertion "Assert failure" failed at ..\..\src\msw\font.cpp(586) in wxNativeFontInfo::SetWeight(): unknown font weight
```

JackMcKew: I believe my drama from above was from not updating WxPython from the `1.0.5-release` branch, I switched branches and tried running the tests there, that worked, switched back and everything is hunky dory. 😄

JackMcKew: I hope I've rebased this right haha 😄 

Let me know if any dramas

chriskiehl: Hmm.. yeah, looks like there may have been some drama in that rebase ^_^ From looking at the commits, I'm guessing you rebased in the wrong 'direction'. 1.0.5 got layered on top of your commits, thus mangling history (and thoroughly confusing Github, which now shows unrelated comments from other entirely different PRs!). 

Easiest way to fix this is probably to grab a fresh copy of 1.0.5, checkout your branch as it existed before you did the rebase, and then do the rebase again. 

If you don't have your changes handy on a backup branch, you can get back to your original state by using git's reflog. 

```
git reflog
```
Here's an example output. This is me rebasing your branch locally on my machine. The bottom sha is the the one before any rebasing took place. 
```
8cdd582 HEAD@{39}: rebase: Add documentation about clear_before_run
8cdbbfe HEAD@{40}: rebase: Add Listbox height option to documentation
b93f4d5 HEAD@{41}: rebase: closes #625 - terminal font and colors not being set correctly
bdbc6c8 HEAD@{42}: rebase: Update french.json
ef3b011 HEAD@{43}: rebase: Fix typo in ISSUE_TEMPLATE.md.
5bb3355 HEAD@{44}: rebase: Ensure OnDropFiles function matches interface
da33e37 HEAD@{45}: rebase: allow control+scrollwheel to zoom rich text console #432
620f797 HEAD@{46}: rebase: add missing return statement in processor.py
e554e38 HEAD@{47}: rebase: closes #592 - clicking close button can cause Gooey to hang
f7ae29a HEAD@{48}: rebase: Do not translate arguments in add_subparsers
833d01e HEAD@{49}: rebase: add test cases for show_stop_warning decorator option
45f4753 HEAD@{50}: rebase: Fix show_stop_warning
1619caa HEAD@{51}: rebase: Resolve all wx warnings and errors and bump version to 4.1.0
285e75f HEAD@{52}: rebase: Added test for parent inheritance behaviour and made "widgets" behave the same as "defaults" for duplicate named arguments.
3bd7e9b HEAD@{53}: rebase: Copy widgets and their configuration from parent parsers. Resolves #608
4e4a661 HEAD@{54}: rebase: Prevent crash when getting dynamic properties
264561f HEAD@{55}: rebase: Add missing chinese translations
7ae42ac HEAD@{56}: rebase: excess space
2f3aeda HEAD@{57}: rebase: typo
78181c2 HEAD@{58}: rebase: closes #593 & #591 - dynamic updates causes slowdowns and loses selection
4cded5d (origin/master, master) HEAD@{59}: rebase: checkout master
29e8fc0 (JackMcKew-1.0.4-release-tqdm-loading-bar) HEAD@{60}: checkout: moving from 1.0.5-release to 1.0.5-release-bk
``` 

You can create a new branch from that pre-rebase sha and be right back where you started

```
git checkout -b you-new-branch-name relevant-sha-from-reflog
```

I suspect 1.0.5 is mangled, so I'd say blow it away locally and then grab it from origin 

```
git branch -D 1.0.5-release 
git fetch 
git checkout -b 1.0.5-release origin/1.0.5-release 
```
That'll put the latest clean version on your machine and swap you to it.  

Finally, checkout your dev branch again 
```
git checkout the-branch-you-created-above-from-the reflog 
```

then rebase from_this_ branch by pointing at 1.0.5-release 

```
git rebase 1.0.5-release 
```

Hopefully that helps!

JackMcKew: Thank you so much for the detailed guide, it helped a lot!

It seemed to be quite a painful git experience but I think we're there, what I had to do was:

1. Create a new branch locally of this PR
2. Hard reset to this sha `git reset --hard 2e56d21`
3. Rebase to `1.0.5-release`
4. Do a force push to push the new local branch onto this PR remote branch `git push -f origin 1.0.4-release-tqdm-loading-bar-rebase:1.0.4-release-tqdm-loading-bar`

All my local testing works, the tests pass, and the PR is only showing 1 commit so I think this is the winner! 😄

chriskiehl: Commits look good! `git` can definitely be a bit of a beast to wrangle. 

Just a few remaining minor issues which popped up during local testing. Should be quick fixes. 

1. The progress bar is displaying on the config screen when you launch Gooey: image

2. If you run with the timing options enabled but _don't_ supply `progress` args, you get errors in `timing.py`

```python
@Gooey(
        program_name="Elapsed / Remaining Timer on Progress in Gooey",
        # progress_regex=r"^progress: (?P<current>\d+)/(?P<total>\d+)$",
        # progress_expr="current / total * 100",
        timing_options = {
                'show_time_remaining':False,
                'show_time_remaining_on_complete':False
        }
        )
```

Specifically [this line](https://github.com/chriskiehl/Gooey/pull/589/files#diff-be5e5016522e8f934cb056dffbf2908aR21) here:
```python
if(prog > 0)
```

Now, obviously, under ideal circumstances, people wouldn't turn on the time estimation feature without also supplying a way for the estimates to be _done_. However, the API allows it, which means we should gracefully handle it and not throw errors. 

Longer term, I'll be tightening up the API validation for all of Gooey's arguments (such that we can easily say A depends on B). But for now, you'll have to just patch around the lack of that capability so that your code doesn't throw errors if the `progress` args haven't been supplied.

JackMcKew: These are both now fixed and I even tidied up the footer label to be a bit cleaner 😄

chriskiehl (approved):

chriskiehl:
This has been merged into release [here](https://github.com/chriskiehl/Gooey/tree/1.0.5-release)! 🎉 🎉

JackMcKew:
Woohoo!!! 🎉🎉🎉🎉

Thank you again for all your help, now I should probably update the PR on gooey examples 🙂

PR #599, opened 2020-07-10
Prevent crash when getting dynamic properties

### LINKED ISSUE #584: poll_external_updates=True not working
I wanted to use the `poll_external_updates` feature, but I can't get it to work. Simply adding that option go Gooey gives errors.

I also tried running the dynamic updates example, but it's giving me the same errors.
https://github.com/chriskiehl/GooeyExamples/blob/master/examples/dynamic_updates.py 

 -  MacOS Mojave 10.14.6 (18G3020)
 -  Python Version  3.6.4
 -  Gooey Version Gooey==1.0.3.1

The error I'm getting is : 

```
 python3 dynamic_updates.py 
  File "dynamic_updates.py", line 122, in <module>
    main()
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/gooey/python_bindings/gooey_decorator.py", line 87, in inner2
    return payload(*args, **kwargs)
  File "dynamic_updates.py", line 55, in main
    args = parser.parse_args()
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/gooey/python_bindings/gooey_parser.py", line 114, in parse_args
    return self.parser.parse_args(args, namespace)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/gooey/python_bindings/gooey_decorator.py", line 82, in run_gooey
    application.run(build_spec)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/gooey/gui/application.py", line 21, in run
    app = build_app(build_spec)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/gooey/gui/application.py", line 30, in build_app
    gapp = GooeyApplication(merge(build_spec, imagesPaths))
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/gooey/gui/containers/application.py", line 70, in __init__
    self.fetchExternalUpdates()
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/gooey/gui/containers/application.py", line 160, in fetchExternalUpdates
    self.buildSpec['encoding']
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/gooey/gui/seeder.py", line 15, in fetchDynamicProperties
    proc = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/subprocess.py", line 709, in __init__
    restore_signals, start_new_session)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/subprocess.py", line 1344, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: "'/Library/Frameworks/Python.framework/Versions/3.6/bin/python3' -u 'dynamic_updates.py' gooey-seed-ui --ignore-gooey": "'/Library/Frameworks/Python.framework/Versions/3.6/bin/python3' -u 'dynamic_updates.py' gooey-seed-ui --ignore-gooey"
```

rotu:
Since `args` is a string, the `Popen` constructor interprets "/usr/bin/python myscript.py" as a path to an executable instead of "/usr/bin/python" with argument "myscript.py".
Adding `shell=True` makes this behave as expected.

Fix #584

### DIFF
--- a/gooey/gui/seeder.py
+++ b/gooey/gui/seeder.py
@@ -12,7 +12,7 @@ def fetchDynamicProperties(target, encoding):
     dynamically generated defaults with which to seed the UI
     """
     cmd = '{} {}'.format(target, 'gooey-seed-ui --ignore-gooey')
-    proc = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    proc = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, shell=True)
     if proc.returncode != 0:
         out, _ = proc.communicate()
         return json.loads(out.decode(encoding))


rotu:
> * The shell variable be configurable by the user rather than blanket set to `True`. There's some existing config available for controlling whether or not things are invoked with `shell`. Checkout the config [here](https://github.com/chriskiehl/Gooey/blob/master/gooey/python_bindings/config_generator.py#L55)

That may be true but the preceding line assumes that target is a string, so it must be executed with shell. Perhaps target should be an array instead?

https://github.com/chriskiehl/Gooey/blob/56a2b8526fadedf2ad7aa8acc543c74f256a7c83/gooey/gui/seeder.py#L14

> * Additionally, you'll need to update this to target the release branch

There are multiple release branches, and it’s not clear which is the “current release branch” per  https://github.com/chriskiehl/Gooey/blob/master/CONTRIBUTING.md#pull-request-process. Maybe the repo default branch should be changed to whichever you want PRs against https://docs.github.com/en/github/administering-a-repository/setting-the-default-branch

chriskiehl:
Thanks for pointing out the confusion with the release branches. 

I've updated the [contributing guidelines](https://github.com/chriskiehl/Gooey/blob/master/CONTRIBUTING.md) to explicitly point at the target branch. Additionally, I've added a [PR Template](https://github.com/chriskiehl/Gooey/blob/master/docs/pull_request_template.md), which'll hopefully clear up any remaining confusion with people opening PRs. 

It'd be nice if Github had a way of specifying branch semantics beyond just `default`. The problem with setting the default branch to an active dev one is that people cloning the project would be grabbing a potentially unstable code base. I prefer master to be a stable mirror of the latest release. 

For the `shell=[configurable]` thing, I don't recall at the top of my head the specific issues which led to the main processor being configurable. Some quick testing on win/osx shows this to be ok, though. How Popen works falls out of my head the second I stop staring at it. 

Can you target this to 1.0.5 so we can merge? 

Thanks!

rotu:
Thanks! I usually expect master to be semi-stable. Anyone who expects a stable release is probably going to be installing off of PyPI instead of via git. I would expect a branch with "release" in the name to be *more* stable than master, rather than less.

> For the `shell=[configurable]` thing, I don't recall at the top of my head the specific issues which led to the main processor being configurable. Some quick testing on win/osx shows this to be ok, though. How Popen works falls out of my head the second I stop staring at it.

I suspect using `shell=False` in gooey is broken in other ways, though without the reason for its existence, I'm not sure why it exists.

PR #601, opened 2020-07-10
Fix warning when dynamically populating dropdown

rotu:
Previously, dynamically populating a dropdown can give this log message: `Gtk-CRITICAL **: 14:51:42.138: gtk_tree_model_iter_nth_child: assertion 'n >= 0' failed`
Now, the warning doesn't happen

### DIFF
--- a/gooey/gui/components/widgets/dropdown.py
+++ b/gooey/gui/components/widgets/dropdown.py
@@ -51,8 +51,9 @@ class Dropdown(TextContainer):
         try:
             yield
         finally:
-            current_at_index = self.widget.GetString(prevSelection)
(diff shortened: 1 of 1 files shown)


chriskiehl:
On which platform does it throw this warning? I wasn't able to reproduce on Windows. 

This did remind me to add another item to the [release checklist](https://github.com/chriskiehl/Gooey/blob/master/docs/releases/release-checklist.md) so that I finally resolve all the warnings. Hopefully it'll help catch similar things in the future. 

The dropdown tests fail with this patch. Can you give those an eyeball?

rotu:
This was on Ubuntu 20.04

PR #609, opened 2020-08-06
Copy widgets and their configuration from parent parsers. Resolves #608

### LINKED ISSUE #608: Widgets are ignored in parent parsers
Argparse supports the notion of parser parents for reusing common arguments, e.g. between subparsers. However, when a GooeyParser is defined with arguments that have custom widgets and is then passed via the parents parameter, those widgets seem to be ignored.

The following code shows the problem, only `b_file` is shown with a FileChooser widget, despite both `a_file` and `b_file` being configured to use this widget. (Tested on Win 10, Gooey 1.0.4, Python 3.8.1)

```python


@Gooey()
def main():
    base_parser = GooeyParser(add_help=False)
    base_parser.add_argument("a_file", widget="FileChooser")

    parser = GooeyParser(parents=[base_parser])
    parser.add_argument("b_file", widget="FileChooser")

    print(parser.parse_args())

if __name__ == "__main__":
    main()
```

image

neonbunny:
Resolves issue #608 based on the sample code in issue.
I'm aware there's no automated tests, but I couldn't work out how to write one for this.

### DIFF
--- a/gooey/python_bindings/gooey_parser.py
+++ b/gooey/python_bindings/gooey_parser.py
@@ -68,6 +68,11 @@ class GooeyParser(object):
         self.__dict__['parser'] = ArgumentParser(**kwargs)
         self.widgets = {}
         self.options = {}
+        if 'parents' in kwargs:
+            for parent in kwargs['parents']:
+                if isinstance(parent, self.__class__):
+                    self.widgets.update(parent.widgets)
+                    self.options.update(parent.options)
 
     @property
     def _mutually_exclusive_groups(self):
--- a/gooey/python_bindings/gooey_parser.py
+++ b/gooey/python_bindings/gooey_parser.py
@@ -93,8 +93,14 @@ class GooeyParser(object):
 
         action = self.parser.add_argument(*args, **kwargs)
         self.parser._actions[-1].metavar = metavar
-        self.widgets[self.parser._actions[-1].dest] = widget
-        self.options[self.parser._actions[-1].dest] = options
+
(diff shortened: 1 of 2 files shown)


gooey/python_bindings/gooey_parser.py
```diff
@@ -68,6 +68,11 @@ def __init__(self, **kwargs):
         self.__dict__['parser'] = ArgumentParser(**kwargs)
         self.widgets = {}
         self.options = {}
+        if 'parents' in kwargs:
```
chriskiehl, line 71: Your example from the PR comments would be a fine test case for this. 

```
def test_parser_parents(): 
    base_parser = GooeyParser(add_help=False)
    base_parser.add_argument("a_file", widget="FileChooser")
    parser = GooeyParser(parents=[base_parser])
    parser.add_argument("b_file", widget="FileChooser")
  
    # assert that widgets and options have been merged here
```

I'd also like to see a test case where it does the 'wrong' thing and overwrites the parent options, which was def unexpected to me that it'd be valid (see my main comment).

neonbunny:
I've added some tests to verify the parent parameter works in the same way as argparse's parent parameter. This opened up a kettle of fish because, as you suggested, argparse is very messy with duplicate named arguments, e.g. it stores them all in a list, but the first one set with a "default" will become the default value for all of them. I believe I have mirrored this logic in Gooey's parser for widgets (just without needing a list).

chriskiehl:
👍 Thanks for the updates! Look good!

PR #618, opened 2020-08-28
Fix #592 - No in Close does not freeze Gooey now

### LINKED ISSUE #592: When the program is running, clicking the close button in the upper right corner will cause the program to freeze.
- [ ] windows10
 - [ ] Python 3.7
 - [ ] Gooey  1.0.4.0.0.0 
 - [ ] Thorough description of problem 
     - [ ]  close the program
     - [ ]  program is hangs
 - [ ] Screenshot 

 - [ ] A minimal code example
Thanks! ^_^ 

```
import sys
import time

@Gooey()
def gooey_app():
    parser = GooeyParser()
    args = parser.parse_args()

    # Do something for a long time
    for i in range(100):
        time.sleep(1)

    return 0

if __name__ == '__main__':
    sys.exit(gooey_app())
```

roopeshvs:
Selecting 'No' in Close Modal caused Gooey to hang.
This was because of control reaching self.Destroy() despite the 'No' selection.
This is fixed simply by adding a boolean flag.

### DIFF
--- a/gooey/gui/containers/application.py
+++ b/gooey/gui/containers/application.py
@@ -145,9 +145,11 @@ class GooeyApplication(wx.Frame):
     def onStopExecution(self):
         """Displays a scary message and then force-quits the executing
         client code if the user accepts"""
+        exit_flag = False
         if not self.buildSpec['show_stop_warning'] or modals.confirmForceStop():
+            exit_flag = True
             self.clientRunner.stop()
-
+        return exit_flag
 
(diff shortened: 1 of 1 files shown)


chriskiehl:
Thanks for the pull! However, I ended up fixing this is a slightly different way: https://github.com/chriskiehl/Gooey/commit/9f2d3726502f5193eaafa07128e0db5564b4a061

PR #633, opened 2020-09-11
Use program_name instead of script file name in macOS menu

jpauwels:
Cosmetic fix of default menu on macOS

Before:
gooey-before

After:
gooey-after

 - [x] All existing tests pass 
 - [n/a] Your bug fix / feature has associated test coverage 
 - [n/a] README.md is updated (if relevant)

### DIFF
--- a/gooey/gui/application.py
+++ b/gooey/gui/application.py
@@ -24,6 +24,8 @@ def run(build_spec):
 
 def build_app(build_spec):
   app = wx.App(False)
+  # use actual program name instead of script file name in macOS menu
+  app.SetAppDisplayName(build_spec['program_name'])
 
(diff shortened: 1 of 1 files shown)


gooey/gui/application.py
```diff
@@ -24,6 +24,7 @@ def run(build_spec):
 
 def build_app(build_spec):
   app = wx.App(False)
+  app.SetAppDisplayName(build_spec['program_name'])
```
chriskiehl, line 27: Could you add a comment here explaining that this for getting MacOS to display things correctly? 

Without it, a year from now, this'd probably be something I accidentally delete while assuming it was redundant or something.

chriskiehl (changes requested):

gooey/gui/application.py
jpauwels: Done

PR #658, opened 2020-12-09
Update options.py

DrStriky:
comparing literal with is not

small fix


chriskiehl:
I merged it into the appropriate branch here: https://github.com/chriskiehl/Gooey/commits/1.0.8-release

PR #659, opened 2020-12-13
german translation additions to match english

brankej:
hey there,

there where multiple buttons without german translations, so i figured to make them match all the english ones.
Thank you!


brankej:
changed to commit into 1.0.7-release

PR #688, opened 2021-04-26
Update portuguese language

eitchtee:
Add missing entries and improve old translation

This PR aims to fix some missing translations on the original file, creating unpleasing results like the one below
image

All while improving current translations to be more natural.


chriskiehl:
lgtm. Thanks for the patch! 👍

PR #709, opened 2021-06-15
Run easy install script if parent script is a windows executable

msochor1:
# Checklist
 - [ ] Works on both Python 2.7 & Python 3.x
 - [x] All existing tests pass 
 - [ ] ~~README.md is updated (if relevant)~~

# Python 2.7 note
Python 2.7 is EoL.  I did try to check if it works on 2.7 and use old-school `.format` but when I created a 2.7 environment and attempted to install Gooey I got the following error:
```ERROR: Could not find a version that satisfies the requirement Pillow>=4.3.0```

So currently I'm not positive Gooey's dependencies are available from 2.7

As such, I'm not going to test python 2.7 and I do apologize.  If this cannot be merged b/c of that, so be it and thank you for your time!

# Issue notes
Fixes related issues https://github.com/chriskiehl/Gooey/issues/455 and https://github.com/chriskiehl/Gooey/issues/649

**Core issue**: for windows install packaged code that is pip installed, there will be an executable .exe file.  When Gooey is run, it cannot find that exe file.  If we instead point to the easy-install file with the suffix `-script.py` then Gooey runs as desired.  Also, depending on install method, there is or is not an easy-install script file to correctly execute the executable.

This PR checks if the script being executed is actually an .exe file.  
If it is an .exe file, then it will check if there is an easy-install script.  
If there is no easy-install script, it will create one in the Scripts directory

It will then run the easy-install script in place of the original target, which will now correctly reference the .exe file.
# System:
```
Windows 10
python 3.6
```

# Manual testing
I have a pip installable package decorated with `@Gooey` at the function which has my Arg parser

1. Test editable install
```
pip uninstall my_package
cd my_package
pip install -e .
```
run your package and confirm it works

2. Test normal install
```
pip uninstall my_package
cd my_package
pip install .
```
run your package and confirm it works

```
pip uninstall my_package
```
confirm file that was created (`my_package-script.py`) is cleaned up by uninstall

### DIFF
--- a/gooey/python_bindings/config_generator.py
+++ b/gooey/python_bindings/config_generator.py
@@ -27,6 +27,41 @@ default_layout = {
 def create_from_parser(parser, source_path, **kwargs):
 
   run_cmd = kwargs.get('target')
+  if os.path.exists('{}.exe'.format(source_path)):
+    # This is installed as a windows executable
+
+    script_path = '{}-script.py'.format(source_path)
+    if not os.path.exists(script_path):
+      # If easy_install script does not exist, create it
+      executable_name = os.path.split(source_path)[-1]
+      easy_install_script = [
+        '#!{}'.format(sys.executable),
+        'import re',
+        'import sys',
+        '__requires__ = "{}"'.format(executable_name),
+        'try:',
+        '    from importlib.metadata import distribution',
+        'except ImportError:',
+        '    try:',
+        '        from importlib_metadata import distribution',
+        '    except ImportError:',
+        '        from pkg_resources import load_entry_point',
+        'def importlib_load_entry_point(spec, group, name):',
+        '    dist_name, _, _ = spec.partition("==")',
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by khavernathy** [comment]
is there a reason this isn't merged? It would fix the same issue for me without needing to change the Gooey source code locally.

**Comment by jtomori** [comment]
@khavernathy if this doesn't get merged then you can point your pip dependency to a fork fixing it.

In my case I've used this: `"Gooey @ git+https://github.com/jtomori/Gooey@issue-455-windows-package-fallback#egg=Gooey"`, which points to my fork with the fix.

I haven't used @msochor1's fork directly, but forked it further to fix some dependency issue.

**Comment by khavernathy** [comment]
This didn't quite fix my error. I got a new/different error now on Windows.
```
importlib.metadata.PackageNotFoundError: No package metadata was found for MY_GUI_PROGRAM
```

Issue #519, opened 2020-01-03
Gooey command to run package as GUI, without having gooey as dependency


pawamoy:
This is a feature request 🙂 

If I understood correctly, we currently need to add Gooey as a dependency of a package to enjoy the generated GUI.

Would it be possible to have a command instead, like `gooey my_package`? Or maybe something more helpful for gooey, like `gooey my_package.cli:main`?

Gooey would "just" need to import the package/module, wrap the specified function into its `Gooey` object, and run it.

The immediate benefit is that we could have generated GUIs for *every Python package that relies on argparse*, instead of requiring the developers of each package to depend on Gooey.

Sorry if this has already been asked (I did not find such issue).

pawamoy:
I closed my PR, maintainer didn't give any news (no problem, just cleaning up). Feel free to take it over.

PR #788, opened 2022-02-03
Conquer the land of arguing snakes

pawamoy:
Implements #519 (draft)

Original PR title is "Add CLI to run any Python program based on argparse", feel free to put it back :joy: 

 - [ ] All existing tests pass: they don't, I have errors like this:
     - `AttributeError: module 'wx' has no attribute 'TaskBarIcon'`
 - [ ] README.md is updated (if relevant)

---

So, this PR basically adds a CLI to Gooey, allowing users to **run any argparse-based Python program from the command line**. I personally think it's awesome. You still need to pass the module path and function name as `module.path:function_name` because I couldn't find a way yet to infer it from the console script name. Indeed, the console script name can be different than the library name. We could try and read the script itself, as it's quite standardized, to find the actual function to import, but it's not bullet-proof either: on Windows such scripts are written as .exe, and I don't think it's easy to read. Maybe there's a way with importlib.metadata.

I went with a simple `sys.argv`, because using argparse in the CLI felt a bit too... recursive? But after all, why not, this would allow to run Gooey on itself, which is kind of awesome too. Imagine typing just "gooey" in the terminal: it opens a simple window with a single input, and a list of detected scripts/libraries/entrypoints (not sure how to do that yet). You type one, hit enter, and a new window pops up, this time with the selected program options and stuff.

### DIFF
--- a/gooey/cli.py
+++ b/gooey/cli.py
@@ -38,4 +38,4 @@ def main(args=None):
     sys.argv = [prog, *args[1:]]
     module = import_module(module_path)
     function = getattr(module, function_name)
-    return Gooey(use_cmd_args=True)(function)
+    return Gooey(function, use_cmd_args=True)()
--- a/gooey/cli.py
+++ b/gooey/cli.py
@@ -28,4 +28,4 @@ def main(args=None):
     sys.argv = [prog, *args[1:]]
     module = import_module(module_path)
     function = getattr(module, function_name)
-    return Gooey(function)
+    return Gooey(use_cmd_args=True)(function)
--- a/gooey/cli.py
+++ b/gooey/cli.py
@@ -2,7 +2,10 @@
 
 import sys
 from importlib import import_module
-from importlib.metadata import entry_points
+try:
+    from importlib.metadata import entry_points
+except ImportError:
+    from importlib_metadata import entry_points
 
 from gooey import Gooey
 
@@ -26,6 +29,9 @@ def main(args=None):
         # the path to a function was passed
         module_path, function_name = args[0].split(':')
         prog = module_path.split('.', 1)[0]
+    else:
+        print('usage: gooey [SCRIPT | MODULE:FUNCTION] [-- SCRIPT_ARGS...]', file=sys.stderr)
+        sys.exit(1)
     if len(args) > 1:
         if args[1] == '--':
             del args[1]
--- a/gooey/cli.py
+++ b/gooey/cli.py
@@ -1,0 +1,18 @@
+'''The command line interface.'''
+
+import sys
+from importlib import import_module
+
+from gooey import Gooey
+
+
+def main(args=None):
+    args = args or sys.argv[1:]
+    module_path, function_name = args[0].split(':')
+    if len(args) > 1:
+        if args[1] == '--':
+            del args[1]
+    sys.argv = ['gooey', *args[1:]]
+    module = import_module(module_path)
+    function = getattr(module, function_name)
+    return Gooey(function)()
--- a/gooey/cli.py
+++ b/gooey/cli.py
@@ -2,17 +2,30 @@
 
 import sys
 from importlib import import_module
+from importlib.metadata import entry_points
 
 from gooey import Gooey
 
 
 def main(args=None):
     args = args or sys.argv[1:]
-    module_path, function_name = args[0].split(':')
+    if not args:
+        script_name = "gooey"
+    else:
+        script_name = args[0]
+    scripts = {script.name: script for script in entry_points().get('console_scripts')}
+    if script_name in scripts:
+        script = scripts[script_name]
+        module_path = script.module
+        function_name = script.attr
+        prog = script_name
+    elif ':' in script_name:
+        module_path, function_name = args[0].split(':')
+        prog = module_path.split('.', 1)[0]
     if len(args) > 1:
         if args[1] == '--':
             del args[1]
-    sys.argv = ['gooey', *args[1:]]
+    sys.argv = [prog, *args[1:]]
     module = import_module(module_path)
     function = getattr(module, function_name)
-    return Gooey(function)()
+    return Gooey(function)
--- a/gooey/cli.py
+++ b/gooey/cli.py
@@ -10,16 +10,20 @@ from gooey import Gooey
 def main(args=None):
     args = args or sys.argv[1:]
     if not args:
+        # without args we assume we want to run gooey on gooey itself
+        # TODO: rewrite using argparse
         script_name = "gooey"
     else:
         script_name = args[0]
     scripts = {script.name: script for script in entry_points().get('console_scripts')}
     if script_name in scripts:
+        # a valid script was passed
         script = scripts[script_name]
         module_path = script.module
         function_name = script.attr
         prog = script_name
     elif ':' in script_name:
+        # the path to a function was passed
         module_path, function_name = args[0].split(':')
         prog = module_path.split('.', 1)[0]
     if len(args) > 1:
--- a/setup.py
+++ b/setup.py
@@ -16,7 +16,7 @@ deps = [
     're-wx>=0.0.9',
     'typing-extensions==3.10.0.2',
     'wxpython>=4.1.0',
-    'dataclasses>=0.8; python_version < "3.7"'
+    'dataclasses>=0.8; python_version < "3.7"',
     'importlib-metadata; python_version < "3.8"',
 ]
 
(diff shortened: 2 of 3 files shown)


gooey/application.py
```diff
@@ -0,0 +1,18 @@
...
+from gooey import Gooey
+
+
+def main(args=None):
+    args = args or sys.argv[1:]
+    module_path, function_name = args[0].split(":")
+    if len(args) > 1:
+        if args[1] == "--":
+            del args[1]
+    sys.argv = ["gooey", *args[1:]]
```
pawamoy, line 15: Need to change "gooey" to the actual script name, though not sure how to infer it from module path + function name yet.

pawamoy, line 15: Also, not sure if it makes sense to accept additional arguments on the command line? Will they be populated in the Gooey window? If not, I'll remove this ability.

pawamoy:
https://user-images.githubusercontent.com/3999221/152335867-d0b404ad-efff-4597-b40b-627d2b030064.mp4

chriskiehl:
+100 points for the sweet demo video! 

I did a bad job of keeping the contributing document updated (Whoops!). 1.0.9 was abandoned in favor of the now released 1.2.0. A bunch of stuff was moved around, so you'll need to target that branch instead (sorry!).

The `application.py` file, for instance, is now `bootstrap.py` and lives in the main `gui` module. There's also a bunch of lingering dead code which you'll have to watch out for, but if you update the PR we can work through any dragons caused by my laziness.

pawamoy:
Damn, it was just there (1.2.0) and I missed it somehow :expressionless: 
No problem, I'll see if I can rebase without too much trouble :slightly_smiling_face:

pawamoy:
OK, branch stuff looks good.

gooey/application.py
chriskiehl: If you enable the `use_cmd_args` option, it'll pass whatever is supplied on the command line into the GUI. So, if you change your call to something like `Gooey(use_cmd_args=True)(function)` ist (should) read and process any additional args.

pawamoy:
I'm able to find the function to import based on the script name thanks to `importlib.metadata.entry_points`.
However the window does not pop up anymore since I rebased on release 1.2.1. Did something changed? Is there a method I need to call on the Gooey instance?

gooey/cli.py
```diff
@@ -0,0 +1,35 @@
+'''The command line interface.'''
+
+import sys
+from importlib import import_module
+from importlib.metadata import entry_points
```
pawamoy, line 5: We probably need to try and import it first, then `importlib_metadata` if it fails.

gooey/cli.py
```diff
@@ -0,0 +1,35 @@
...
+
+from gooey import Gooey
+
+
+def main(args=None):
+    args = args or sys.argv[1:]
+    if not args:
+        # without args we assume we want to run gooey on gooey itself
+        # TODO: rewrite using argparse
```
pawamoy, line 14: Running Gooey on itself requires rewriting this function with an argparse parser.

gooey/cli.py
```diff
@@ -0,0 +1,35 @@
...
+    if script_name in scripts:
+        # a valid script was passed
+        script = scripts[script_name]
+        module_path = script.module
+        function_name = script.attr
+        prog = script_name
+    elif ':' in script_name:
+        # the path to a function was passed
+        module_path, function_name = args[0].split(':')
+        prog = module_path.split('.', 1)[0]
```
pawamoy, line 28: Probably need an else clause to handle errors and print useful error messages.

setup.py
```diff
@@ -17,6 +17,7 @@
     'typing-extensions==3.10.0.2',
     'wxpython>=4.1.0',
     'dataclasses>=0.8',
+    'importlib-metadata; python_version < "3.8"',
```
chriskiehl, line 20: Ah, whoops! Thanks for this fix!

gooey/cli.py
```diff
@@ -0,0 +1,41 @@
...
+    else:
+        print('usage: gooey [SCRIPT | MODULE:FUNCTION] [-- SCRIPT_ARGS...]', file=sys.stderr)
+        sys.exit(1)
+    if len(args) > 1:
+        if args[1] == '--':
+            del args[1]
+    sys.argv = [prog, *args[1:]]
+    module = import_module(module_path)
+    function = getattr(module, function_name)
+    return Gooey(use_cmd_args=True)(function)
```
chriskiehl, line 41: Sorry, I led you astray with my other comment (didn't actually test it before commenting!). 

```
Gooey(function, use_cmd_args=True)()
```

or, because Python is weird: 

```
Gooey(use_cmd_args=True)(function)()
```

gooey/cli.py
```diff
@@ -0,0 +1,41 @@
+'''The command line interface.'''
+
+import sys
+from importlib import import_module
+try:
+    from importlib.metadata import entry_points
```
chriskiehl, line 6: Explain a bit about these different versions?

pawamoy: `importlib.metadata` is included in the standard library since Python 3.8 only.
Since you support lower versions, the `importlib-metadata` package is required (it's a simple backport).
Therefore, we must handle both cases in the code since the import path is not the same (`importlib.metadata` vs. `importlib_metadata`).

gooey/cli.py
chriskiehl: I'll leave it up to you ^_^ 

I'd be fine going sans additional configuration for the initial release of this -- let it "bake" a bit and see if people need the functionality. On the flip side, if you want to throw a parser in there so people can control things that's fine, too. You may be able to even generate a options for all of the [params](https://github.com/chriskiehl/Gooey/blob/master/gooey/python_bindings/parameters.py) directly from the type definition by looking through its `__annotations__` (in theory).

chriskiehl:
Re: tests: Give a pull on 1.2.1 and the tests should run for you. There were previously a lot of broken paths because I'm an IDE scrub and PyCharm was patching all the paths for me. 

Re: merging the feature: you'll need to also do the less fun stuff and update the docs ^_^ We should update the [Usage](https://github.com/chriskiehl/Gooey#usage) section of the README with a small example. Additionally, depending on how much it takes to explain, you can expand the "How To" either later with a specific section in the README, or with lengthier guidance in the `docs/` directory.

pawamoy:
Better, no path-related errors anymore, though it seems I have an incompatible wxPython/rewx version? All tests fail with `AttributeError: module 'wx' has no attribute 'TaskBarIcon'`. Any hint? :smile:

chriskiehl:
>Do I have an incompatible system lib?

Nay, that was a bug in rewx. I pushed a fix. If you upgrade the dep things should start working again.

pawamoy:
Still no luck :sweat_smile: 

```console
% python -m unittest
.Traceback (most recent call last):
  File "/media/data/dev/Gooey/.venv/lib/python3.9/site-packages/wx/core.py", line 2217, in OnPreInit
    self.InitLocale()
  File "/media/data/dev/Gooey/gooey/tests/__init__.py", line 60, in InitLocale
    locale.setlocale(locale.LC_ALL, lang)
  File "/home/pawamoy/.basher-packages/pyenv/pyenv/versions/3.9.9/lib/python3.9/locale.py", line 610, in setlocale
    return _setlocale(category, locale)
locale.Error: unsupported locale setting
[1]    158722 segmentation fault (core dumped)  python -m unittest
```

I'm on Arch by the way :trollface:

pawamoy:
Maybe you could setup a test suite in GitHub CI in the meantime.

pawamoy:
Something like this:

```yaml
# .github/workflows/ci.yml
name: ci

jobs:
  tests:
    strategy:
      matrix:
        os:
        - ubuntu-latest
        - macos-latest
        - windows-latest
        python-version:
        - "3.7"
        - "3.8"
        - "3.9"
        - "3.10"
        - "3.11-dev"

    runs-on: ${{ matrix.os }}

    steps:
    - name: Checkout
      uses: actions/checkout@v2    

    - name: Set up Python
      uses: actions/setup-python@v2
      with:
        python-version: ${{ matrix.python-version }}

    - name: Install project and dependencies
      run: pip install .

    - name: Run the test suite
      run: python -m unittest
```

pawamoy:
Hi @chriskiehl, how are you? Just coming back to ask you how do you see this going forward? Would you like to setup CI using GitHub Actions? I can send a PR with the contents above, though it'll probably need some tinkering (installing system libs, etc.). Or maybe you'd simply prefer I finish the TODOs so we can merge?

PR #798, opened 2022-03-18
Add missing German translations

TobiasRosskopf:


Ric16Fr (approved):
I was just about to do the same, thanks for adding

PR #818, opened 2022-05-07
fix installation 1.2.0-alpha on python > 3.7

namujae:
Hi, Thank you for the good library.

this patch will fix the installation error for 1.2.0-alpha.

the error came from dataclasses pkg requirements on setup.py. it was embeded in default python from 3.7.


chriskiehl:
Thanks for trying out the alpha! And thanks for the patch!

PR #821, opened 2022-05-18
Improve german translation

S1SYPHOS:
This commit adds missing translations & improves existing ones :tada:


gooey/languages/german.json
```diff
@@ -1,37 +1,50 @@
 {
   "browse": "Durchsuchen",
   "cancel": "Abbrechen",
+  "checkbox_label": "Aktiviere",
+  "choose_colour": "",
```
smokephil, line 5: Here the translation is missing. What do you think of "Farbe auswählen"

S1SYPHOS: That works for me 😀

PR #886, opened 2023-07-01
Change "is not" to "!=" when checking literals

BlueishTint:
In a most widget files `is not` is used to check the error against an empty string (`''`). This doesn't work all the time [to my knowledge](https://stackoverflow.com/questions/2209755/python-operation-vs-is-not), so it is best to replace it with "!=", which is an equality check.

**Example**

Before:
```py
def syncUiState(self, state: FormField):  # type: ignore
        self.widget.setValue(state['value'])  # type: ignore
        self.error.SetLabel(state['error'] or '')
        self.error.Show(state['error'] is not None and state['error'] is not '')
```
After:
```py
def syncUiState(self, state: FormField):  # type: ignore
        self.widget.setValue(state['value'])  # type: ignore
        self.error.SetLabel(state['error'] or '')
        self.error.Show(state['error'] is not None and state['error'] != '')
#       --------------------------------------------------------------^^----
```
Checklist:
 - [ ] Commits have been squashed and includes the relevant issue number

 Squashing is not necessary (only one commit) and there is no relevant issue.
 - [x] All existing tests pass

N/A
 - [ ] README.md is updated (if relevant)

N/A

### DIFF
--- a/gooey/gui/components/widgets/bases.py
+++ b/gooey/gui/components/widgets/bases.py
@@ -183,7 +183,7 @@ class TextContainer(BaseWidget):
     def syncUiState(self, state: FormField):  # type: ignore
         self.widget.setValue(state['value'])  # type: ignore
         self.error.SetLabel(state['error'] or '')
-        self.error.Show(state['error'] is not None and state['error'] is not '')
+        self.error.Show(state['error'] is not None and state['error'] != '')
 
 
     def getValue(self) -> t.FieldValue:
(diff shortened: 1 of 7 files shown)


gooey/gui/components/widgets/bases.py
```diff
@@ -183,7 +183,7 @@ def getUiState(self) -> t.FormField:
     def syncUiState(self, state: FormField):  # type: ignore
         self.widget.setValue(state['value'])  # type: ignore
         self.error.SetLabel(state['error'] or '')
-        self.error.Show(state['error'] is not None and state['error'] is not '')
+        self.error.Show(state['error'] is not None and state['error'] != '')
```
elad-eyal, line 186: isn't it enough to write

```
self.error.Show(bool(state['error']))
```
?

KianKhadempour: That seems to work as well, assuming we want other falsy values to equate to false as well (such as 0).

Issue #801, opened 2022-04-04
Title color changes with os dark mode and cannot be fixed


s5409693:
Hello,

Gooey is a great help for me to deploy code to non-programmers, but some people have their OS in dark mode. Gooey picks up the dark mode in some widgets, but not in all, so I would like to prescribe the font and background color of every part of a Gooey application, but the program title and program description cannot be changed.

- OS: Debian 11
- Python version: 3.8.8
- Gooey version: 1.0.8.1
- Problem description: I have noticed that when my OS (Debian 11) is in dark mode, some of the widgets adopt the dark theme and some don't, making the application look very strange. So I tried to prescribe the font and background color of all widgets, which worked fine. For the program title and description, however, I have found no way to alter the font color. If it is possible and I just have not found the way how to do it, it would be nice to add this to the documentation. Please find some example code below
- Expected behavior: Either (a) the program title can be prescribed or (b) if this is not possible, at least the program title does not change depending on the user's OS settings.
- Actual behavior: I have not found a way to prescribe the color and when I change gnome-tweaks -- appearance -- themes -- applications from Adwaita to Adwaita-dark, the title color changes from black to white.
- Code example:
```
import gooey

@gooey.Gooey(
        header_color = '#ff0000',       # tried, but has no effect
        header_font_color = '#ff0000',  # tried, but has no effect
        header_bg_color = '#0000ff',    # this works fine
        )
def main():
    parser = gooey.GooeyParser(description='Some test parser')
    parser.add_argument('input_filename', type=str, widget='FileChooser')
    args = parser.parse_args()

if __name__ == "__main__":
    main()
```
- Appearance with Adwaita
![gooey_light](https://user-images.githubusercontent.com/76200839/161538110-4121d6b7-f415-429e-8410-b819106ba93d.png)
- Appearance with Adwaita-dark
![gooey_dark](https://user-images.githubusercontent.com/76200839/161538161-2a561898-f68c-491b-8686-ee767c0b11f3.png)

Thank you and best regards
Paul Seibert

Issue #707, opened 2021-06-14
Use system colors as defaults (dark mode fix)


spyoungtech:
By default, Gooey chooses the default color scheme for apps.

In dark mode, this can make apps a bit difficult to use, particularly on MacOS. For example, notice the fields are difficult to see and the button text is not readable:

![image](https://user-images.githubusercontent.com/15212758/121958066-16c74780-cd18-11eb-9c6c-77a99b85c213.png)

I would suggest that, by default, Gooey should use the system colors, which will account for dark mode and other user-configured OS options. I.E. using [`wx.SystemSettings.GetColour`](https://wxpython.org/Phoenix/docs/html/wx.SystemSettings.html#wx.SystemSettings.GetColour). This way, applications use color schemes according to user system preferences.

spyoungtech:
Using system colors, dark mode might look something like this:


![image](https://user-images.githubusercontent.com/15212758/121967443-3b75ec00-cd25-11eb-9a11-65ae77fdec58.png)

lifekaizen:
I noticed that too! From the Widget Example on Mac:
<img width="151" alt="Screen Shot 2021-06-16 at 9 19 12 PM" src="https://user-images.githubusercontent.com/7156025/122331526-d5869180-cee9-11eb-957a-1b78542ea0a0.png">
_In Light Mode_
<img width="148" alt="Screen Shot 2021-06-16 at 9 19 46 PM" src="https://user-images.githubusercontent.com/7156025/122331532-d7505500-cee9-11eb-8f06-fe58860b0342.png">
_In Dark Mode_

On Mac, the state can be determined by a subprocess output:
```python
>>> # in dark mode
>>> wx.SystemSettings.GetAppearance().IsDark()
False
>>> import subprocess
>>> subprocess.Popen(cmd, stdout=subprocess.PIPE,stderr=subprocess.PIPE, shell=True).communicate()
(b'Dark\n', b'')
>>> # in light mode
>>> wx.SystemSettings.GetAppearance().IsDark()
False
>>> subprocess.Popen(cmd, stdout=subprocess.PIPE,stderr=subprocess.PIPE, shell=True).communicate()
(b'', b'2021-06-16 21:25:45.850 defaults[40272:12750955] \nThe domain/default pair of (kCFPreferencesAnyApplication, AppleInterfaceStyle) does not exist\n')
```

ustun:
@enxyz 

>  wx.SystemSettings.GetAppearance().IsDark()

[...]

Issue #543, opened 2020-04-11
Bad defaults on macos dark mode


LeoFCardoso:
Please consider fix default color values for macos "dark mode".
Example: https://github.com/LeoFCardoso/pdf2pdfocr/issues/17

chriskiehl:
fill out the issue template, please

LeoFCardoso:
Sorry, please consider the template.

 - [ ] macOS Catalina 10.15.4 
 - [ ] Python 3.7.6 
 - [ ] Gooey 1.0.3 (from pip) 
 - [ ] Thorough description of problem:
     - [ ] Expected Behavior = Good color defaults on macOS dark mode.
     - [ ] Actual Behavior = Bad default colors in dark mode. Light mode is ok.
 - [ ] A minimal code example -> https://github.com/LeoFCardoso/pdf2pdfocr/blob/master/pdf2pdfocr_gui.py
 - [ ] Screenshot -> https://github.com/LeoFCardoso/pdf2pdfocr/issues/17

MireskaWillow:
Same with the new 1.0.4.

 - [ ] OS: **Mac OS Catalina 10.15.5**
 - [ ] Python Version: **Python 3.7.7**
 - [ ] Gooey Version: **Gooey 1.0.4.0.0.0**
 - [ ] Thorough description of problem: **Button color defaults are wrong in dark mode, cannot read label. **
     - [ ] Expected Behavior: Button text color is different from button background color.
     - [ ] Actual Behavior: Button text color is same as button background color.
 - [ ] A minimal code example -- preferably copy/pastable in the issue itself (less time figuring out how to run your code == more time debugging!) 
 - [ ] Screenshot (if visual quirk) : [https://i.imgur.com/jAWTBfN.png](https://i.imgur.com/jAWTBfN.png)
 - [ ] Anything else you may think will be helpful: None

```#!/usr/bin/env python3
from gooey import Gooey
import argparse

@Gooey
def main():
    parser = argparse.ArgumentParser()
    parser.add_argument("What… is the air-speed velocity of an unladen swallow?")
    args = parser.parse_args()

if __name__ == "__main__":
    main()```

dldinternet:
I have done some debugging and investigation on where the color defaults for wxPython widgets are that break dark mode.
I have experimented with the widgets_demo.py of GooeyExamples. 
Here is what I have so far:
`'''
@author: Chris
Created on Dec 21, 2013
 __          __  _
 \ \        / / | |
[...]

PR #891, opened 2023-09-05
Feat/better dark mode

nicolasbraun:
# What it si about

This PR is about trying to fix the dark modes issues reported mostly on Mac OS in issues #543 , #707 , #801.

# How?
- Add options to customise the header text colors
- Adapt default based on the current system Dark mode status

# Limitations
- Does not handle mode switch during execution which can cause some text colors to automatically change (mostly the one not specifically set up and using WX defaults) while the background does not.

# Note
Dis not really not again which branch to PR since the 1.0.8 seems to be the current stable release. It should probably be merged as an hotfix. Otherwise I'll try to fix conflicts with 1.21


k-gruenberg:
This one should definitely be merged! It's such a nice tool/idea but pretty useless if you can't see anything with with light text on a light background (and there being no way to alter the font color)...

dg-pb:
Would be great if this was merged

sbillinge:
@dg-pb @k-gruenberg @nicolasbraun I will look at this later today.  @nicolasbraun this branch has become conflicted.  Please could you resync it and repush?  I didn't check what the conflict was but hopefully it is something small.

sbillinge:
@nicolasbraun thanks so much for this PR!

I am about to work on a  new release so I think it is better if can merge it into the `1.2.1-release` branch, especially as this changes the pin of the wx version in requirements and I don't really want to do a hotfix that does that, just in case.

@nicolasbraun to make it easier to merge, please could you remove the style edits to top of the README.  Of course, it is ok to add the edits that explain the new feature.  If you like the style changes, feel free to put them on a separate PR.

I don't have easy access to a Mac so this would be way easier to review if you could grab screenshots of a gui and paste them into the comments. In particular, I am interested to see a more or less side-by-side comparison of what the GUI looks like now and after the change when it is using the default colors.

nicolasbraun:
Hello @sbillinge 

I haven't touched gooey for a year as the related project to my PR was canceled.
I'll try to find a bit a time to make a dummy project and post some screenshots

sbillinge:
@cadenmyers13 would you be interested in trying to generate those screenshots to help out @nicolasbraun ?

dg-pb:
Screenshots look great! 👍

sbillinge:
thanks @nicolasbraun a great service.

johan456789:
What's the status of this PR? The screenshots look great

dg-pb:
Good news! Could someone tell me when 1.2.1 is planned to be in PyPI?

sbillinge:
@johan456789 I can work on that in the next week.

sbillinge:
This work had to be reverted because it broke tests on linux.  It is not straightforward to reopen this PR without nasty things happening to the git history, but, note to selves, we need to revisit this and make these edits again in a way that doesn't break linux.  Please see #936 for discussion on this.

Issue #925, opened 2025-03-02
Why are DropDown values Editable after being selected?


FuturityUK:
Hi, 

When I select a choice from the DropDown menu, I'm then able to edit the value. 

When I click the "Start" button, it fails if the values have been edited because it doesn't match one of the choice restrictions. 

This also happens in the widget_demo.py DropDowns.

Is this expected behaviour? Is there a way to force it to be Read Only do that the user can't mess with the values after they select one?

Thanks

Neil

sbillinge:
This doesn't strike me as desired behavior though there may be a reason that @chriskiehl knows about.

PR #945, opened 2026-03-14
widgets: change dropdown style to readonly

simonsbench:
The Dropdown widget's wx.ComboBox creation used the wx.CB_DROPDOWN style flag,
which allows selection but also makes the selection editable. wx.CB_READONLY
allows selection but without editing, so the selection will always be in the
enumerated choice. See: https://docs.wxpython.org/wx.ComboBox.html#wx-combobox

Fixes: #925


simonsbench:
DROPDOWN_options_are_editable
READONLY_options_are_not_editabe

simonsbench:
See the before and after.

Any guidance on running the tests (I can't seem to find a doc? Running discovery with unittest yields many failures)

simonsbench:
This is against 1.0.9, which I'm guessing is the appropriate place for fixes of 1.0.8 to go

Issue #773, opened 2022-01-04
1.0.9 breaks support for non-Windows OSs


gberardinelli:
Hello,

I'm on the `1.0.9-release` branch looking into adding a feature, but this branch seems to have broken support for non-Windows operating systems in this commit:  https://github.com/chriskiehl/Gooey/commit/13a0c9e182b2787810073b10ed8f156b2ede59bf#diff-c9a4afaa360d6d71bcf32ad1253f3840ad139f43e88733ba3a11b3927294d9fdL87-L92

This is because `subprocess.CREATE_NEW_PROCESS_GROUP` is a Windows only constant, as shown in the [docs](https://docs.python.org/3/library/subprocess.html#windows-constants) and confirmed in the [cpython source](https://github.com/python/cpython/blob/f1a58441eea6b7788c64d03a80ea35996301e550/Lib/subprocess.py#L78-L88).

The GUI comes up initially, but hitting "start" results in an exception:

```
Traceback (most recent call last):
  File "/home/gberardinelli/Scripts/Gooey/gooey/gui/processor.py", line 75, in run
    creationflags=subprocess.CREATE_NEW_PROCESS_GROUP)
AttributeError: module 'subprocess' has no attribute 'CREATE_NEW_PROCESS_GROUP'
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "/home/gberardinelli/venv/gooey/lib/python3.6/site-packages/wx/core.py", line 3407, in <lambda>
    lambda event: event.callable(*event.args, **event.kw) )
  File "/home/gberardinelli/Scripts/Gooey/gooey/gui/containers/application.py", line 93, in onStart
    self.clientRunner.run(self.buildCliString())
  File "/home/gberardinelli/Scripts/Gooey/gooey/gui/processor.py", line 81, in run
    creationflags=subprocess.CREATE_NEW_PROCESS_GROUP
AttributeError: module 'subprocess' has no attribute 'CREATE_NEW_PROCESS_GROUP'
```

I can work on a PR but it may be some time before I can test on Windows, so I figured I'd raise it here first.

For completeness, I'm on Linux Mint 19.3 / Python 3.6 but of course this should apply to anything that's not Windows.

Cheers

chriskiehl:
Ah, thanks for the report. I'll have to revisit. ^_^ 

[...]

PR #946, opened 2026-03-14
widgets: change dropdown style to readonly

simonsbench:
The Dropdown widget's wx.ComboBox creation used the wx.CB_DROPDOWN style flag,
which allows selection but also makes the selection editable. wx.CB_READONLY
allows selection but without editing, so the selection will always be in the
enumerated choice. See: https://docs.wxpython.org/wx.ComboBox.html#wx-combobox

Fixes: #925

### DIFF
--- a/gooey/gui/components/widgets/dropdown.py
+++ b/gooey/gui/components/widgets/dropdown.py
@@ -21,7 +21,7 @@ class Dropdown(TextContainer):
             # str conversion allows using stringyfiable values in addition to pure strings
             value=str(default),
             choices=[str(default)] + [str(choice) for choice in self._meta['choices']],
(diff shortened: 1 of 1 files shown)


simonsbench:
DROPDOWN_options_are_editable
READONLY_options_are_not_editabe

See dropdown options as selectable but editable before the style fix, and then only selectable after the fix.

simonsbench:
(I found this discussion about 1.0.9 (https://github.com/chriskiehl/Gooey/issues/773), and it sounds like it isn't a good release candidate. I'll issue bug fixes that aren't feature improvements against 1.0.8-release, as this makes releasing them easy.)

PR #951, opened 2026-09-03
Update dependencies and versions for latest Python compatibilities

cadenmyers13:
In preparation for the patch release to include Python versions 3.8-3.14. 

I locally installed this into a python 3.14 env and ran `Gooey` using `diffpy.labpdfproc` and it worked.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -6,7 +6,7 @@ from setuptools import setup, find_packages
 with open('README.md') as readme:
     long_description = readme.read()
 
-version = '1.0.8.1'
+version = '1.0.8.2'
 
 deps = [
     'Pillow>=4.3.0',
--- a/setup.py
+++ b/setup.py
@@ -1,5 +1,4 @@
 """Script for setuptools."""
-import sys
 from setuptools import setup, find_packages
 
 
--- a/setup.py
+++ b/setup.py
@@ -29,7 +29,6 @@ setup(
     install_requires=deps,
     include_package_data=True,
     python_requires='>=3.8',
-    dependency_links = ["http://www.wxpython.org/download.php"],
     classifiers = [
         'Development Status :: 4 - Beta',
         'Intended Audience :: Developers',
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -31,15 +28,20 @@
     packages=find_packages(),
     install_requires=deps,
     include_package_data=True,
-    dependency_links = ["http://www.wxpython.org/download.php"],
```
cadenmyers13, line 34: `dependency_links` has been non-functional since 2019 and is listed as deprecated: https://setuptools.pypa.io/en/latest/deprecated/dependency_links.html

cadenmyers13:
@sbillinge ready for review

setup.py
```diff
@@ -1,24 +1,20 @@
...
     'psutil>=5.4.2',
-    'colored>=1.3.93',
-    'pygtrie>=2.3.3'
+    'colored>=1.3.93,<2',
+    'pygtrie>=2.3.3',
+    'wxpython>=4.1.0',
 ]
 
-if sys.version[0] == '3':
-    deps.append('wxpython>=4.1.0')
```
cadenmyers13, line 19: This is no longer necessary since we're dropping python 2 in the next release. We just add `wxpython` as a dependency.

sbillinge:
This looks good to me, though if we are dropping python 2 in this version it will not be backwards compatible and so we would need to do a mid or top level version bump, not a patch one.  

There is a case to be made to do a patch release without dropping python 2 and then drop python 2 in the next `1.2.x` release that has all the API changes?

IF we just remove the python 2 work from this PR, is that all that is needed to keep it backwards compatible or there are more changes in the code that we have already done that would need to be reverted?

sbillinge:
I think ignore that last comment. This only goes back to 3.8 and I saw from the README that we already only claim that it works for Python 3 so this is not a change in this release.

PR #953, opened 2026-09-10
doc & support: Support python 3.9+ and change documentation

cadenmyers13:
Changing the documentation to reflect support for python 3.9+ and making this change in setup.py


cadenmyers13:
@sbillinge ready for review

